Issue #48, opened 2015-03-02
Login on heruko redirects to localhost


brettminnie:

eddiejaoude:
Yep, I need to create another App on Github for Production (heroku) & move the matching parameters for that to config. Thanks for the reminder :+1:

eddiejaoude:
Created a **Github OAuth App** for `Dev` & one for `Production` - the `Production` one is via ENV variables.

eddiejaoude:
Still occurring. I will investigate why the **ENV** variables are not being picked up.

brettminnie:
did we not have this issue before with heruko and env variables?

eddiejaoude:
I don't think so. The DB ENV vars all work fine. I will look into it in the morning.

eddiejaoude:
One will have to run `> composer install` locally to reset **parameters.yml**.

eddiejaoude:
Still occurring. Even though, the redirect url in the url params is correct :cry:

eddiejaoude:
Deleting OAuth Apps on Github & starting again

eddiejaoude:
Creating `Prod` one first. But still trying to use `Dev` parameters.

eddiejaoude:
Right, I have gotten to the bottom of this issue. ENV params do not override the **.dist** template, it is a known issue & has been around for a while https://github.com/symfony/symfony/issues/7555

The solution is to move the **dev** parameters to their own **dev** file.

brettminnie:
Nicely done

Bit of a pain that

eddiejaoude:
Deployment failed...

```
Generating optimized autoload files
       Creating the "app/config/parameters.yml" file
       
       
                                                                                                
         [Symfony\Component\DependencyInjection\Exception\ParameterCircularReferenceException]  
         Circular reference detected for parameter "client.id" ("client.id" > "client.id"). 
```

Hmm

eddiejaoude:
This is why I like to use my SymfonyQuickStart, all this is already setup. Good experience / learning curve I guess with Symfony.

Fixed the above **Circular Reference**. Onto next issue :)

eddiejaoude:
It does fail now, but doesn't show user logged in either.

eddiejaoude:
The log...

```
[...]

PR #50, opened 2015-03-03
[#48] Renamed OAuth parameter names

eddiejaoude:
As described in symfony docs
http://symfony.com/doc/current/cookbook/configuration/external_parameter
s.html

Not sure why the DB parameters work slightly differently :confused:


eddiejaoude:
One will have to run `> composer install` locally to reset **parameters.yml**.

brettminnie:
Possibly due to the library being used

PR #59, opened 2015-03-04
Improve docs and fix bugs on vagrant setup

Netsensia:
The README.md file is now good enough to get you up-and-running properly. Fixed NFS problem that prevented apache2 from being installed. Fixed various problems that stopped the system coming up 100% accurately when following the README file. Requires VMWare.


eddiejaoude:
Why `bum` for directory? What does it stand for?

brettminnie:
Yeah, that may require a rename. Though I did chuckle

Netsensia:
The bum was a mistake - I wanted to put the bum up a level as it was the name for my pillar directory, but I already had a directory called pillar, so I chose bum. Now deleted from PR.

brettminnie:
Wow... Just wow ;-)

PR #83, opened 2015-03-14
Hotfix/81 mock user error

eddiejaoude:


app/config/security.yml
```diff
@@ -4,7 +4,6 @@ security:
     # http://symfony.com/doc/current/book/security.html#encoding-the-user-s-password
     encoders:
         Symfony\Component\Security\Core\User\User: plaintext
-        DashboardHub\Bundle\AppBundle\Entity\User: plaintext
```
brettminnie: you going to leave this as plaintext?

brettminnie:
happy to merge if ready

app/config/security.yml
eddiejaoude: I don't even think that it is needed because we don't store anything at all regarding the user information as this is done via OAuth.

eddiejaoude: We can't remove it, it is required by the OAuth bundle, but I guess it doesn't do the encoding because it doesn't store the information itself & leaves it to the OAuth Service (in this case Github)

eddiejaoude:
Thanks, lets do it! :+1:

brettminnie:
Merged

PR #130, opened 2015-04-09
Add php5.6 support

brettminnie:
Travis config updated to have 5.6 support and experimental 7.0 support as an allowed failure

Added fast_finish flag to ensure if one build fails the rest are terminated and the whole build flagged as a failure

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: php
 
 php:
   - 5.5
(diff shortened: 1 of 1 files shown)


eddiejaoude:
CI passed, Scrutiniser seems to have crashed. But will merge as not part of source code. CI is the important part here.

Thanks for PR :+1:

PR #167, opened 2015-04-27
Use HTTPS to download composer with Travis

Netsensia:
Noticed that non HTTPS download stopped working causing Travis builds to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
   - mysql -e "create database IF NOT EXISTS dashboardhub;" -uroot
 
 before_script:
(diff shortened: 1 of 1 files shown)


eddiejaoude:
Thanks for PR @Netsensia .

That other failure is one that is intermittent - I have solve in another branch, so should stop happening in the near future. 

I will trigger a rebuild again.

eddiejaoude:
All Green now :)

eddiejaoude:
I will need to do the same to there repos - why couldnt they do a redirect from `http` to `https`?

Issue #184, opened 2016-01-19
Implement cache for MVP


simoncasey:
- [x] Configure Spring caching abstraction
- [x] Use simple cache implementation eg Google Guava
- [ ] Create beans for interacting with cache throughout the app

simoncasey:
Enabling the cache in spring only envolved adding a dependency and annotation. Once we have something to cache, we can explore more

eddiejaoude:
:+1: Sounds good. Keep it simple for now :smile:

PR #189, opened 2016-01-20
#184 add caching dependency, google guava and enable spring cache

simoncasey:


src/main/resources/application.yml
```diff
@@ -5,6 +5,10 @@ spring:
   resources:
     chain:
       enabled: true
+  cache:
```
eddiejaoude: What is this for?

PR #198, opened 2016-01-28
Feature/197 oauth service

eddiejaoude:


src/main/java/io/dashboardhub/pipelinedashboard/config/SecurityConfig.java
```diff
@@ -18,6 +19,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
+@Profile("!test")
```
simoncasey: is this needed now that we have moved the test mock auth into test package?

eddiejaoude: I think so, because we don't want this one to run during `test` mode. Unless you see a better way?

PR #200, opened 2016-01-30
Feature/199 font awesome

eddiejaoude:
- [ ] @simoncasey can you double check the security exception for `fonts` in path `src/main/resources/static/fonts` is required?


simoncasey:
@eddiejaoude any reason why we arent using a cdn hosted fontawesome?

eddiejaoude:
> ...using a cdn hosted fontawesome...

I found that it was slower, but maybe only during development. We can always change it back if required i the future

Issue #195, opened 2016-01-27
Abstract away pages using PageObject model pattern


eddiejaoude:
* [ ] Page Object model
* [ ] Open / Close browser for each feature

PR #205, opened 2016-01-31
Feature/185 audit user login

eddiejaoude:
* [x] Use `lastLoggedIn` field instead of Login table - **Envers** will audit these changes
* [ ] ~~Close browser after each feature is run `@After`~~ This will go into separate PR

screenshot 2016-02-04 07 55 25


src/main/java/io/dashboardhub/pipelinedashboard/controller/AccountController.java
```diff
@@ -0,0 +1,24 @@
...
+
+@Controller
+public class AccountController {
+
+    @Autowired
+    private UserService userService;
+
+    @RequestMapping(value = {"/profile"}, method = RequestMethod.GET)
+    public String profile(Model model) {
+        User user = this.userService.findByCurrentUsername();
```
simoncasey: why this.userService? there isnt any other UserService in scope ...

src/main/java/io/dashboardhub/pipelinedashboard/domain/User.java
```diff
@@ -23,10 +26,10 @@
     private String username;
 
     private String name;
+    private String email;
 
-//    @ManyToMany
-//    @JoinTable(name="users_projects")
-//    private Set<Project> projects;
+    @OneToMany(cascade = CascadeType.ALL, fetch = FetchType.EAGER)
```
simoncasey: this is gonna kill performance if there is a login recorded each time the user logs in. This set will become massive and be eagerly fetched each time the User record is. Suggest reversing the relationship as knowledge of login history only really valid in admin/support/debug situations.

src/main/java/io/dashboardhub/pipelinedashboard/listener/AuthenticationListener.java
```diff
@@ -20,5 +21,9 @@ public void onApplicationEvent(final InteractiveAuthenticationSuccessEvent event
         if (user == null) {
             this.userService.save(new User(event.getAuthentication().getName()));
         }
+
+        user = this.userService.findByUsername(event.getAuthentication().getName());
+        user.getLogins().add(new Login());
```
simoncasey: is there a benefit to having this logged in db over log file? alternatively, could we incorporate this into Envers auditing by adding a lastLoggedIn field to the User object?

src/test/java/io/dashboardhub/pipelinedashboard/definitions/GuestDefinitions.java
```diff
@@ -21,6 +21,5 @@ public void I_try_to_access_a_secure_page() {
     @Then("^I get redirected to the login page$")
     public void I_get_redirected_to_the_login_page() {
         Assert.assertEquals(withBaseUrl("/login"), driver.getCurrentUrl());
-        driver.close();
```
simoncasey: where does the driver get closed now?

src/main/java/io/dashboardhub/pipelinedashboard/controller/AccountController.java
eddiejaoude: Habit of using `this` for class scope. Is that bad?

src/main/java/io/dashboardhub/pipelinedashboard/domain/User.java
eddiejaoude: `EAGER` is probably not right, it was a question I had for you as it would not work if wasn't like this. 

You are right though, may be a **bidirectional** relationship is better, because the user can see their login history and the **admin** can see all login history and for what user

src/main/java/io/dashboardhub/pipelinedashboard/listener/AuthenticationListener.java
eddiejaoude: Log file will only be looked at when something goes wrong, so not good for Audit purpose.

Could it be incorporated into *Envers*...oh I see using, `lastLoggedIn` field... I LIKE IT!!!

src/test/java/io/dashboardhub/pipelinedashboard/definitions/GuestDefinitions.java
eddiejaoude: Good question, because I share the functionality between **contexts**, it closes it at the end only. This might present an issue in the future, but won't know until then

src/main/java/io/dashboardhub/pipelinedashboard/domain/User.java
eddiejaoude: Not an issue if I change it to the `lastLoggedIn` field in the **User** class and then *Envers* to audit the changes.

src/main/java/io/dashboardhub/pipelinedashboard/controller/AccountController.java
simoncasey: yeah it's rarely necessary

eddiejaoude: Ok, I will remove the `this` where not required - which is probably everywhere :smile:

eddiejaoude: Done :+1:

src/test/java/io/dashboardhub/pipelinedashboard/definitions/GuestDefinitions.java
eddiejaoude: I have added it to this ticket, which I do shortly https://github.com/DashboardHub/PipelineDashboard/issues/195

PR #208, opened 2016-02-04
Feature/207 update profile

eddiejaoude:
* [x] Update Name & Email
* [x] Validation Errors

screenshot 2016-02-04 07 55 46

screenshot 2016-02-04 07 55 35


src/main/java/io/dashboardhub/pipelinedashboard/listener/AuthenticationListener.java
```diff
@@ -7,6 +7,8 @@
 import org.springframework.security.authentication.event.InteractiveAuthenticationSuccessEvent;
 import org.springframework.stereotype.Component;
 
+import java.util.Date;
```
simoncasey: avoid using Date class wherever possible. Use the Java8 time/date api instead - eg LocalDateTime

src/main/java/io/dashboardhub/pipelinedashboard/domain/User.java
```diff
@@ -19,14 +20,15 @@
...
+    @Size(min = 5, max = 255)
     private String name;
 
-//    @ManyToMany
-//    @JoinTable(name="users_projects")
-//    private Set<Project> projects;
+    @Email
+    private String email;
+
+    private String lastLoggedIn;
```
simoncasey: String??? use a LocalDateTime with the @Temporal annotation

eddiejaoude: I started with a Date object, but had issues with the result coming back from Thymeleaf. I wanted to get your feedback on it.

What is @Temporal?

TeMPOraL: It's me, and I get a notification anytime someone writes this phrase on GitHub :disappointed:.

eddiejaoude: Oh, your username is the same as a Spring boot / Java feature?! Yeah that can't be fun :disappointed:

TeMPOraL: Yup. All lands in my mailbox. I get to discover interesting projects pretty much every other day though. Like this one, which I think I even recognize from HN :).

eddiejaoude: :smile:

simoncasey: haha that is a unfortunate side effect of github user tagging - i bet their are lots of users that get a similar deal with all the java annotation. @eddiejaoude @-temporal is a java persistence annotation, not specific to spring, but to JPA

eddiejaoude: @simoncasey is going to do as part of Envers work

PR #209, opened 2016-02-06
Feature/174 projects

eddiejaoude:
* [x] #208 is merged into `prototype-v0.3` then `prototype-v0.3` onto here, this **PR** will reduce in size


src/main/java/io/dashboardhub/pipelinedashboard/service/ProjectService.java
```diff
@@ -1,12 +1,71 @@
...
+
+        Project existingProject = findByUid(project.getUid());
+        if (existingProject != null) {
+            if (existingProject.getUser() != user) {
+                throw new PermissionDeniedException("Not owner of Project");
+            }
+
+            existingProject.setName(project.getName());
+            existingProject.setDescription(project.getDescription());
+            existingProject.setIsPrivate(project.getIsPrivate());
```
eddiejaoude: @simoncasey Is there a better way to do the above? Update existing `Project` with *updates*

src/main/java/io/dashboardhub/pipelinedashboard/service/ProjectService.java
```diff
@@ -1,12 +1,71 @@
...
+
+    public Iterable<Project> findAllByPublic() {
+        return this.projectRepository.findAllPublic();
+    }
+
+    public Project findByUid(String uuid) {
+        return this.projectRepository.findByUid(uuid, getCurrentUsername());
+    }
+
+    public Boolean delete(Project project) {
```
simoncasey: why does this need to return anything?

eddiejaoude: Fair point. I will update

src/main/java/io/dashboardhub/pipelinedashboard/service/ProjectService.java
simoncasey: This is roughly correct i think

src/main/java/io/dashboardhub/pipelinedashboard/service/ProjectService.java
```diff
@@ -1,12 +1,71 @@
...
 public class ProjectService {
 
     @Autowired
-    ProjectRepository projectRepository;
+    private ProjectRepository projectRepository;
+
+    @Autowired
+    private UserService userService;
+
+    private String getCurrentUsername() {
```
simoncasey: Why is this duplicated within UserService?

src/main/java/io/dashboardhub/pipelinedashboard/service/ProjectService.java
eddiejaoude: Ok cool. I was not sure if there was an easier / magical way to do it.

src/main/java/io/dashboardhub/pipelinedashboard/service/ProjectService.java
eddiejaoude: I could do it as an `Advice`?

src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProjectDefinitions.java
```diff
@@ -3,10 +3,38 @@
...
 import cucumber.api.java.en.When;
 import io.dashboardhub.pipelinedashboard.PipelinedashboardApplicationTests;
+import io.dashboardhub.pipelinedashboard.domain.Project;
+import io.dashboardhub.pipelinedashboard.domain.User;
+import io.dashboardhub.pipelinedashboard.repository.ProjectRepository;
+import io.dashboardhub.pipelinedashboard.repository.UserRepository;
 import org.junit.Assert;
+import org.springframework.beans.factory.annotation.Autowired;
 
 public class ProjectDefinitions extends PipelinedashboardApplicationTests {
```
simoncasey: I'm confused here. I thought we were using cucumber purely for acceptance tests (full application tests via selenium). Are we now using the same for integration tests (which this class looks like)??

eddiejaoude: This class is seeding the database with dummy data for the Scenario.

No reason why we can't use it for Integration tests too? We do this on other projects.

src/main/java/io/dashboardhub/pipelinedashboard/service/ProjectService.java
simoncasey: yup, or just call a method on UserService as you have it in scope at line 18

src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProjectDefinitions.java
simoncasey: ok cool

src/main/java/io/dashboardhub/pipelinedashboard/service/ProjectService.java
eddiejaoude: :+1:

eddiejaoude: Actually an **Advice** already exists for the **CurrentUser**, it just needs updating with the full User object https://github.com/DashboardHub/PipelineDashboard/blob/prototype-v0.3/src/main/java/io/dashboardhub/pipelinedashboard/controller/advice/CurrentUserAdvice.java

eddiejaoude: I took your advice (no pun intended) and used the **UserService** for now

PR #214, opened 2016-03-27
204 entity auditing

simoncasey:
Added mysql config (look in application.yml for settings). Currently drops/creates db on each launch. Example revision history lookup at /profile/history.


src/main/java/io/dashboardhub/pipelinedashboard/config/ApplicationConfig.java
```diff
@@ -0,0 +1,15 @@
+package io.dashboardhub.pipelinedashboard.config;
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+
+@Configuration
+public class ApplicationConfig {
+
+//    @Profile("!test")
```
eddiejaoude: Is this meant to be commented out?

src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProjectDefinitions.java
```diff
@@ -10,6 +10,8 @@
 import org.junit.Assert;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import java.util.UUID;
```
eddiejaoude: Is this needed?

src/main/java/io/dashboardhub/pipelinedashboard/domain/Project.java
```diff
@@ -8,17 +11,15 @@
...
+@ToString(callSuper = true)
 @Entity
+@EntityListeners(AuditingEntityListener.class)
 @Table(uniqueConstraints = @UniqueConstraint(name = "uid_idx", columnNames = "uid"))
-public final class Project {
-
-    @Id
-    @GeneratedValue(strategy = GenerationType.AUTO)
-    private Long id;
+public class Project extends BaseEntity {
```
eddiejaoude: `final` got removed, is there a reason?

src/main/java/io/dashboardhub/pipelinedashboard/domain/Project.java
```diff
@@ -28,8 +29,13 @@
...
 
     @NotNull
-    private Boolean isPrivate = false;
+    private Boolean isPrivate;
 
     @ManyToOne(fetch = FetchType.EAGER)
     private User user;
+
+    public Project() {
+        this.isPrivate = false;
```
eddiejaoude: Is it better to do it on the constructor that on the class property?

eddiejaoude:
We need **feature** tests to cover the changes.

src/main/java/io/dashboardhub/pipelinedashboard/config/ApplicationConfig.java
simoncasey: good point. missed this when sending PR. I'll cancel the pr

src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProjectDefinitions.java
simoncasey: This is a leftover from when I was using proper UUIDs. It'll be needed when we use them, but otherwise it can just be removed.

src/main/java/io/dashboardhub/pipelinedashboard/config/ApplicationConfig.java
eddiejaoude: Don't cancel the PR! Just update :smile:

src/main/java/io/dashboardhub/pipelinedashboard/domain/Project.java
simoncasey: merge issue. Still not sure why we want to mark all the entities as final though ...

src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProjectDefinitions.java
eddiejaoude: I would remove for now, to keep it cleaner. We can add back into the PR that requires them

src/main/java/io/dashboardhub/pipelinedashboard/domain/Project.java
simoncasey: I think so. It's easier to read IMO

simoncasey:
regarding feature tests - there were no user-facing changes made so it's a bit difficult to do feature tests for it (not including the revision history as that is a just an example that will not remain). Also, I'd like to pair with you on the use of the cucumber-style testing and to review our testing strategy (when we use unit, integration, feature)

src/main/java/io/dashboardhub/pipelinedashboard/domain/Project.java
eddiejaoude: It's similar to `private` vs `public` reasons, make it *locked down* and *open up* as required. 

DI is better than Inheritance

src/main/java/io/dashboardhub/pipelinedashboard/domain/Project.java
eddiejaoude: Ok

eddiejaoude:
Why close it? Can just add to the **PR** :confused:

eddiejaoude:
Cucumber is not just for UI tests, it is just the abstraction. The FeatureContext determines the context, here we use the Context to check the DB https://github.com/DashboardHub/PipelineDashboard/blob/prototype-v0.3/src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProjectDefinitions.java#L22

eddiejaoude:
I did see a html page added, but even if there wasn't, the feature something like this:

```cucumber
  Scenario: Updating user entity should be audited 
    Given I have a User
    When I update the name
    Then I see a new audit entry
```

simoncasey:
ok. I added a feature to test the auditing and amended as per the little comments above. Also, changed the default (and therefore test) profile back to using H2 and changed travis to drop the mysql requirement. 

when running locally, if you want to use mysql, run the app in the dev profile.

src/main/java/io/dashboardhub/pipelinedashboard/config/ApplicationConfig.java
```diff
@@ -0,0 +1,14 @@
+package io.dashboardhub.pipelinedashboard.config;
```
eddiejaoude: Application config is very generic, would it not be better to rename it to **DateConfig**?

src/main/java/io/dashboardhub/pipelinedashboard/config/CurrentTimeDateTimeService.java
```diff
@@ -0,0 +1,11 @@
+package io.dashboardhub.pipelinedashboard.config;
+
+import java.time.ZonedDateTime;
+
+public class CurrentTimeDateTimeService implements DateTimeService {
```
eddiejaoude: If this class is called `CurrentTimeDateTimeService`, why is it in the **config** directory?

src/main/java/io/dashboardhub/pipelinedashboard/config/AuditingDateTimeProvider.java
```diff
@@ -0,0 +1,20 @@
+package io.dashboardhub.pipelinedashboard.config;
+
+import org.springframework.data.auditing.DateTimeProvider;
+
+import java.util.Calendar;
+import java.util.GregorianCalendar;
+
+public class AuditingDateTimeProvider implements DateTimeProvider {
```
eddiejaoude: This file ends in **Provider**, should this not go into a **Provider** directory?

src/main/java/io/dashboardhub/pipelinedashboard/config/DateTimeService.java
```diff
@@ -0,0 +1,8 @@
+package io.dashboardhub.pipelinedashboard.config;
+
+import java.time.ZonedDateTime;
+
+public interface DateTimeService {
```
eddiejaoude: **Service** in the `config` directory?

src/main/java/io/dashboardhub/pipelinedashboard/config/UsernameAuditorAware.java
```diff
@@ -0,0 +1,19 @@
+package io.dashboardhub.pipelinedashboard.config;
+
+import org.springframework.data.domain.AuditorAware;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+
+public class UsernameAuditorAware implements AuditorAware<String> {
```
eddiejaoude: I am not sure about the naming of the end of the file. Would it not be better placed in an `Auditor` directory?

src/main/java/io/dashboardhub/pipelinedashboard/config/UsernameAuditorAware.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public String getCurrentAuditor() {
+
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+
+        if (authentication == null || !authentication.isAuthenticated()) {
+            return null;
```
eddiejaoude: A note for the future, use the **NullObject** Pattern here

simoncasey: Hmm, might be difficult here since we have no control over what the api returns. That said, this is a copy/paste from the tutorials on the net

eddiejaoude: No worries, this is low priority for now :smile: ... more food for thought

src/main/resources/templates/account/history.html
```diff
@@ -0,0 +1,32 @@
...
+            Last logged in at <strong th:text="${user.lastLoggedIn}"></strong>
+        </div>
+    </div>
+
+    <div class="row">
+        <div class="card">
+            <table>
+                <tr th:each="revision : ${revisions}">
+                    <td th:text="${revision.getRevisionDate()}"></td>
+                    <td th:text="${revision.getEntity().toString()}"></td>
```
eddiejaoude: Also need a Feature test for this page

src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProfileDefinitions.java
```diff
@@ -1,7 +1,61 @@
...
+
+import java.lang.reflect.Field;
+import java.util.List;
 
 public final class ProfileDefinitions extends PipelinedashboardApplicationTests {
 
+    @Autowired
+    private UserRepository userRepository;
+
+    @Then("^There is a new revision for my User$")
```
eddiejaoude: No capital required otherwise would read `Then There...` but should be `Then there...`

src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProfileDefinitions.java
```diff
@@ -1,7 +1,61 @@
...
+    public void There_is_a_new_audit_entry_for_my_user() throws Exception {
```
eddiejaoude: No capital for method name

src/test/resources/io/dashboardhub/pipelinedashboard/definitions/profile/profile.feature
```diff
@@ -25,3 +25,19 @@ Feature: User can update their profile
...
+    Given I am logged in
+    When I go /profile
+      And I fill in the field name with testname
+      And I fill in the field email with test@email.com
+      And Submit the form form-profile
+    Then I go /profile
+      And I fill in the field name with updatedtestname
+      And I fill in the field email with updatedtest@email.com
+      And Submit the form form-profile
+    Then There is a new revision for my User
```
eddiejaoude: :+1:

src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProfileDefinitions.java
```diff
@@ -1,7 +1,61 @@
...
+    private UserRepository userRepository;
+
+    @Then("^there is a new revision for my User$")
+    public void there_is_a_new_audit_entry_for_my_user() throws Exception {
+        User currentUser = userRepository.findByUsername("TestUser");
+        Revision<Integer, User> latestRevision = userRepository.findLastChangeRevision(currentUser.getId());
+        Assert.assertTrue(latestRevision.getRevisionNumber() > 1);
+    }
+
+    @And("^the latest revised value for (.*) is (.*)$")
```
eddiejaoude: If under `Then` section, should be `@Then`. `And` is to generic and could be used in any section.

src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProfileDefinitions.java
```diff
@@ -1,7 +1,61 @@
...
+    public void the_latest_revised_value_for(String fieldName, String value) throws Exception {
+        checkRevisionFieldValue(fieldName, value, 0);
+    }
+
+    @And("^the previous revised value for (.*) is (.*)$")
```
eddiejaoude: If under `Then` section, should be `@Then`. `And` is to generic and could be used in any section.

eddiejaoude: I have changed it

src/test/java/io/dashboardhub/pipelinedashboard/definitions/ProfileDefinitions.java
eddiejaoude: I have changed it

eddiejaoude:
@simoncasey if you are happy with my additions, I am happy to merge the Pull Request

simoncasey:
@eddiejaoude sure go ahead and merge :)

eddiejaoude:
Cool. Lets do it! Merging...

PR #863, opened 2018-05-19
feat(terms-conditions): #597 Added terms and conditions page

### LINKED ISSUE #597: T&Cs page
If first issue just make holding page and don’t worry about the content.

### Requirements

**API**
- [ ] Task One
- [ ] Task Two

**UI**
- [ ] Task One
- [ ] Task Two

eddiejaoude:
closes #597 

### Notes

- [ ] API: Task One (eg. updated `deployed` model )
- [ ] UI: Task Three (eg. added validation rules on create environment page)
- [ ] Documentation: Updated accordingly

### DIFF
--- a/web/src/app/app.routes.ts
+++ b/web/src/app/app.routes.ts
@@ -27,6 +27,7 @@ import { MonitorsResolver } from './environments/monitors/monitors.resolve';
 import { MonitorsAddComponent } from './environments/monitors/add/monitors-add.component';
 import { PingsResolver } from './environments/monitors/pings.resolve';
 import { LogoutComponent } from './auth/logout/logout.component';
+import { TermsConditionsComponent } from './legal/terms-conditions/terms-conditions.component';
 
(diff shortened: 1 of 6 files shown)


web/src/app/app.routes.ts
```diff
@@ -43,6 +44,11 @@ const routes: Routes = [
         pathMatch: 'full',
         component: HelpComponent,
       },
+      {
+        path: 'terms-and-conditions',
+        pathMatch: 'full',
```
andrew-cunliffe: Not really an issue perhaps its good to have but the path will work fine without setting this, its more useful on things that are ambiguous and you want to do a redirect in certain cases

eddiejaoude: Thanks for the feedback @andrew-cunliffe 👍  . That is something @chowbertilla and I did talk about. I kept it in there for consistency

PR #883, opened 2018-05-21
feat(environments): #872 strip forward slash from url

### LINKED ISSUE #872: Strip / on environment url
### Requirements

**API**

In the `environment` model, we should be able to define a `setter` method that strips off a `/`, but we will need to use this new method on the `create` and `update`.

- [x] Write a test for creating an environment with a url that ends in a `/` and check that the API removes it `api/tests/features/environments/create.feature` (this will initially fail until the code is written)
- [x] Write a test for updating an environment with a url that ends in a `/` and check that the API removes it `api/tests/features/environments/update.feature` (this will initially fail until the code is written)
- [ ] Create a setter method on the environment model `api/src/models/environment.js`
   - After the schema definition, probably on line `104`
   - The method will look something like this...

```javascript
schema.methods.setLink = function(link) {
    console.log(link); // check for ending with '/' if so, then remove
};
```

- [x] In the setter method above, write code to remove `/` at the end of the link if it exists

More information on `setter` methods in the ORM docs https://dynamoosejs.com/api#methods

Now we need to make use of the new setter method in the `create` and `update` environment

- [ ] In the create `api/src/environments/create.js` remove line `14` and add `environmentModel.setLink(data.link);` to line `29` just before the `save`
- [ ] In the update `api/src/environments/update.js` in the `switch` statement in `case '/link':` also use the setter method `environment.setLink(item.value);` and put a break afterwards, should look something  like...

```javascript
case '/link':
   environment.setLink(item.value);
   break;
```

chowbertilla:
closes #872 

### Notes

- [ ] API: Task One (eg. updated `deployed` model )
- [ ] UI: Task Three (eg. added validation rules on create environment page)
- [ ] Documentation: Updated accordingly


api/tests/features/environments/monitors/create.feature
```diff
@@ -11,6 +11,20 @@ Feature: Create Monitor
     When I make a "POST" request to "/environments/e0000001-0000-0000-0000-000000000000/monitors"
     Then the status code should be 403
 
+  Scenario: Checking URL format conforms when creating an environment
+    Given I am logged in
+    When I make a "POST" request to "/environments" with:
+        | field   | value                |
+        | link    | "http://example.com" |
+    Then the status code should be 200
+    And should have a field "link" with "http://example.com"
```
eddiejaoude: This test should go in the `api/tests/features/environments/create.feature` file

api/tests/features/environments/monitors/create.feature
```diff
@@ -11,6 +11,20 @@ Feature: Create Monitor
...
+
+    Then(/^should have a field "([^"]*)" with "([^"]*)"$/, function(field, value) {
+      return this.validate(this.response.body[field], value);
+    });
+    And should have a field "id" with "UUID"
```
eddiejaoude: Lines `22` to `25` are not needed 🤓

api/tests/features/environments/update.feature
```diff
@@ -34,3 +34,11 @@ Feature: Update my environment
     And should have a field "id" with "e0000001-0000-0000-0000-000000000000"
     And should have a field "owner" with "RwbFXSyjZ6F3HJ3Qc5CGUNMdSX8f3m79@clients"
     And should have a field "title" with "UpdateEnvironment A"
+
+  Scenario: Checking URL format conforms when updating an environment
+    Given I am logged in
+    When I make a "POST" request to "/environments" with:
```
eddiejaoude: For update, the method type is `PATCH`.

`GET` = without ID is for a list
`GET` = with ID is for a single item
`POST` = create
`PATCH` = update

I hope that helps 🤓

PR #899, opened 2018-05-22
feat(pings): #893 incomplete conditional for no pings

### LINKED ISSUE #893: Environment total pings should include invalid
### Requirements

- total pings on environment to include invalid (API or UI?), the monitor list has invalid and valid, these should be added together on the environment page

screenshot 2018-05-22 13 08 50

- widen the space for the url on the monitor list page

screenshot 2018-05-22 13 09 01

chowbertilla:
closes #893 

@eddiejaoude So the issue with this is: I have the two `<span>` blocks for each case. However, I could not for the life of me find the file where the ngif resides so that I can change the if then statement accordingly. Please help!

### DIFF
--- a/web/src/app/environments/list/environments-list.component.html
+++ b/web/src/app/environments/list/environments-list.component.html
@@ -75,13 +75,6 @@
 
 <div class="mat-content" layout-padding flex>
   <mat-card>
-    <mat-card-title>Welcome to DashboardHub! To get you started...</mat-card-title>
-    <mat-card-subtitle>This is where I want to insert the markdown text from welcome-message.html</mat-card-subtitle>
-  </mat-card>
-</div>
-
-<div class="mat-content" layout-padding flex>
-  <mat-card>
     <mat-card-title>Environments</mat-card-title>
     <mat-card-subtitle>Recent activity</mat-card-subtitle>
     <mat-divider></mat-divider>


web/src/app/environments/view/environments-view.component.html
```diff
@@ -130,8 +130,11 @@ <h3 matLine>{{ environment.latestPing.url }}</h3>
           <span class="mat-caption text-right" flex="30" layout-align="row" *ngIf="!environment.latestPing.isValid">
             <p>No pings</p>
           </span>
+          <span class="mat-caption text-right" flex="30" layout-align="row" *ngIf="environment.latestPing.isValid">
```
eddiejaoude: With both conditions (this one and the one above), we need to check if there are any pings available, in the variable environment.pings. I think this was me originally making a copy and paste error

web/src/app/environments/view/environments-view.component.html
```diff
@@ -130,8 +130,11 @@ <h3 matLine>{{ environment.latestPing.url }}</h3>
           <span class="mat-caption text-right" flex="30" layout-align="row" *ngIf="!environment.latestPing.isValid">
```
eddiejaoude: Here is the other `*ngIf`

web/src/app/environments/view/environments-view.component.html
```diff
@@ -130,8 +130,11 @@ <h3 matLine>{{ environment.latestPing.url }}</h3>
...
             <p>No pings</p>
           </span>
+          <span class="mat-caption text-right" flex="30" layout-align="row" *ngIf="environment.latestPing.isValid">
+            <p>{{ monitor.pings.valid }} valid pings</p>
+            <p>{{ monitor.pings.invalid }} invalid pings</p>
```
eddiejaoude: On this page there is no `monitor` variable (double check in the network tab), but as it is at the environment level, the variable used here is `environment`, so the `valid` and `invalid` would be `environment.pings.valid` and `environment.pings.valid`. I can explain more on our next pairing session 🤓

Issue #891, opened 2018-05-22
Welcome message on landing page


sarajaoude:
### Requirements

Text box (alert info box?) above the environment list. (between the top boxes and the environment list)

- welcome message
- this page is showing you...
- description of what DH is
- link to the quickstart guide
- login link

eddiejaoude:
PR has been merged

PR #900, opened 2018-05-22
feat(docs): #891 add new welcome-message file

chowbertilla:
@eddiejaoude The issue here is that I would like to NOT hard code the welcome message onto the homepage, so I decided to create a separate file called welcome-message.md - however, I wasn't sure how to redirect environments-list.component.html to look at welcome-message.md - had a look at stackoverflow but I think the answers were overcomplicating the issue and I couldn't figure it out :(

### DIFF
--- a/web/src/app/environments/list/environments-list.component.html
+++ b/web/src/app/environments/list/environments-list.component.html
@@ -76,7 +76,13 @@
 <div class="mat-content" layout-padding flex>
   <mat-card>
     <mat-card-title>Welcome to DashboardHub! To get you started...</mat-card-title>
-    <mat-card-subtitle>This is where I want to add in the text from welcome-message.md</mat-card-subtitle>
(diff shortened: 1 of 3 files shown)


eddiejaoude:
Let’s keep things simple for now and hard code it in the homepage, the card component also has a mat-content html tag that it could go into. 

It’s looking good 👍

web/src/app/environments/list/environments-list.component.html
```diff
@@ -73,6 +73,13 @@
   </div>
 </div>
 
+<div class="mat-content" layout-padding flex>
+  <mat-card>
+    <mat-card-title>Welcome to DashboardHub! To get you started...</mat-card-title>
+    <mat-card-subtitle>This is where I want to add in the text from welcome-message.md</mat-card-subtitle>
+  </mat-card>
+</div>
```
eddiejaoude: Looks good 👍

eddiejaoude:
Looks good 👍

PR #1194, opened 2019-04-24
chore(feat): #1192, Fixed private project should not access publicly

### LINKED ISSUE #1192: Private projects should not be accessible publicly
Server-side is protected, [firebase rules]( https://github.com/DashboardHub/PipelineDashboard/blob/firebase-mvp/firestore.rules#L20)

```gherkin
Scenario: Public projects are accessible by everyone
   Given I am not logged in
   When I access a public project
   Then I can see the project details

Scenario: Private projects are not accessible to guest or anyone other than the owner
   Given I am not logged in
   When I access a private project
   Then I am redirect to the homepage

Scenario: Private projects are only accessible by the owners
   Given I am logged in
   When I access a private project
   Then I can see the project details
```

webkhushboo:
closes #1192 

### Notes

- [ ] Fixed user not logged in should not access private project details using route
- [ ] Fixed user not logged in should not edit private project details using route
- [ ] Fixed user admin and logged in should be able to edit the public projects
- [ ] Fixed user not admin and logged in should not be able to edit the public projects using route

Kindly review and let me know in case of any question

### DIFF
--- a/web/src/app/projects/edit/edit.component.ts
+++ b/web/src/app/projects/edit/edit.component.ts
@@ -2,23 +2,27 @@ import { Component, OnInit } from '@angular/core';
 import { FormGroup, FormBuilder, Validators } from '@angular/forms';
 import { Router, ActivatedRoute } from '@angular/router';
 import { MatSnackBar } from '@angular/material';
-import { ProjectService } from '../../../services/project.service';
-import { catchError } from 'rxjs/operators';
-import { ProjectModel, RepositoryModel } from '../../../models/index.model';
 import { Subscription } from 'rxjs';
+import { catchError } from 'rxjs/operators';
+
+// Dashboard hub models
+import { ProjectModel, RepositoryModel } from '../../../models/index.model';
+
+// Dashboard hub services
+import { AuthenticationService } from '../../../services/authentication.service';
+import { ProjectService } from '../../../services/project.service';
 
 @Component({
     selector: 'dashboard-projects-edit',
     templateUrl: './edit.component.html',
 })
 export class EditProjectComponent implements OnInit {
-
     private projectSubscription: Subscription;
     private saveSubscription: Subscription;
-    private repositorySubscription: Subscription;
     public uid: string;
     public projectForm: FormGroup;
     public repositories: RepositoryModel[] = [];
+    public project: ProjectModel = new ProjectModel();
 
     constructor(
         private route: ActivatedRoute,
@@ -26,6 +30,7 @@ export class EditProjectComponent implements OnInit {
         private form: FormBuilder,
         private snackBar: MatSnackBar,
         private projectService: ProjectService,
+        private authService: AuthenticationService
     ) {
         this.uid = this.route.snapshot.paramMap.get('uid');
     }
@@ -39,7 +44,14 @@ export class EditProjectComponent implements OnInit {
 
         this.projectSubscription = this.projectService
             .findOneById(this.uid)
-            .subscribe((project: ProjectModel) => this.projectForm.reset(project));
+            .subscribe((project: ProjectModel) => {
+                    this.project = project;
+                    this.projectForm.reset(project);
+                    // Added check for not authenticated user and private project details
+                    if ((project.type === 'private' && !this.authService.isAuthenticated) || !this.isAdmin()) {
+                        this.router.navigate(['/']);
+                    }
+            });
     }
 
     save(): void {
@@ -49,6 +61,10 @@ export class EditProjectComponent implements OnInit {
                 catchError((error: any): any => this.snackBar.open(error.message, undefined, { duration: 5000 }))
             )
             .subscribe(() => this.router.navigate(['/project', this.uid]));
+    }
+
+    isAdmin(): boolean {
+        return this.project.access.admin.includes(this.authService.profile.uid);
     }
 
     ngDestroy(): void {
--- a/web/src/app/projects/edit/edit.component.ts
+++ b/web/src/app/projects/edit/edit.component.ts
@@ -6,7 +6,7 @@ import { Subscription } from 'rxjs';
 import { catchError } from 'rxjs/operators';
 
 // Dashboard hub models
-import { ProjectModel, RepositoryModel } from '../../../models/index.model';
+import { RepositoryModel, ProjectModel } from '../../../models/index.model';
 
 // Dashboard hub services
 import { AuthenticationService } from '../../../services/authentication.service';
@@ -22,7 +22,6 @@ export class EditProjectComponent implements OnInit {
     public uid: string;
     public projectForm: FormGroup;
     public repositories: RepositoryModel[] = [];
-    public project: ProjectModel = new ProjectModel();
 
     constructor(
         private route: ActivatedRoute,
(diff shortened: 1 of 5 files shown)


web/src/app/projects/edit/edit.component.ts
```diff
@@ -39,7 +44,14 @@ export class EditProjectComponent implements OnInit {
 
         this.projectSubscription = this.projectService
             .findOneById(this.uid)
-            .subscribe((project: ProjectModel) => this.projectForm.reset(project));
+            .subscribe((project: ProjectModel) => {
+                    this.project = project;
+                    this.projectForm.reset(project);
+                    // Added check for not authenticated user and private project details
+                    if ((project.type === 'private' && !this.authService.isAuthenticated) || !this.isAdmin()) {
+                        this.router.navigate(['/']);
```
eddiejaoude: Is it possible to put this check in a `guard`?

web/src/app/projects/view/view.component.ts
```diff
@@ -28,57 +34,73 @@ export class ViewProjectComponent implements OnInit {
...
     }
 
     ngOnInit(): void {
         this.projectSubscription = this.projectService
             .findOneById(this.project.uid)
-            .subscribe((project: ProjectModel) => this.project = project);
+            .subscribe((project: ProjectModel) => {
+                this.project = project;
+                // Added check for not authenticated user and private project details
+                if (project.type === 'private' && !this.authService.isAuthenticated) {
```
eddiejaoude: Need to check if they are the owner also?

web/src/app/projects/edit/edit.component.ts
eddiejaoude: Or resolve?

webkhushboo: I have checked if we can create guard or not . 
We are checking project is public or private based upon the uid and then we are retrieving the project details from firebase but within guard we can not retrieve the through route without navigating to it. so we have to put a check in component itself for this.

web/src/app/projects/view/view.component.ts
webkhushboo: For viewing the public projects we need not to check if they are owner also. Any one can see the project right but in case of edit project only owner is allowed to edit the public and private projects.

web/src/resolvers/edit-project.resolver.ts
```diff
@@ -0,0 +1,28 @@
...
+import { ProjectService } from '../services/project.service';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class EditProjectResolver implements Resolve<boolean> {
+
+    constructor(private authService: AuthenticationService,
+                private projectService: ProjectService,
+                private router: Router) {
```
eddiejaoude: I think the formatting of this similar to above would be easier to read...

```typescript
constructor(
   private a: b,
   private c: d
) {
```

web/src/app/projects/view/view.component.ts
```diff
@@ -28,57 +34,69 @@ export class ViewProjectComponent implements OnInit {
...
     ) {
         this.project.uid = this.route.snapshot.paramMap.get('uid');
     }
 
     ngOnInit(): void {
         this.projectSubscription = this.projectService
             .findOneById(this.project.uid)
-            .subscribe((project: ProjectModel) => this.project = project);
+            .subscribe((project: ProjectModel) => {
+                this.project = project;
```
eddiejaoude: This can go back to being a single line

eddiejaoude:
Nice work. I left some small comments, but looks good 👍

web/src/app/projects/view/view.component.ts
webkhushboo: Sure will fix. Thanks :)

web/src/resolvers/edit-project.resolver.ts
webkhushboo: Sure !!

web/src/app/projects/view/view.component.ts
eddiejaoude: 👍

web/src/app/projects/view/view.component.ts
```diff
@@ -40,9 +40,7 @@ export class ViewProjectComponent implements OnInit {
     }
 
     ngOnInit(): void {
-        this.projectSubscription = this.projectService
-            .findOneById(this.project.uid)
-            .subscribe((project: ProjectModel) => {
+        this.projectSubscription = this.projectService.findOneById(this.project.uid).subscribe((project: ProjectModel) => {
```
eddiejaoude: sorry, I didn't meant the whole line, I just mean the return in the subscribe

```typescript
.subscribe((project: ProjectModel) => this.project = project)
```

web/src/resolvers/edit-project.resolver.ts
```diff
@@ -10,9 +10,11 @@ import { ProjectService } from '../services/project.service';
...
 export class EditProjectResolver implements Resolve<boolean> {
 
-    constructor(private authService: AuthenticationService,
-                private projectService: ProjectService,
-                private router: Router) {
+    constructor(
+        private authService: AuthenticationService,
+        private projectService: ProjectService,
+        private router: Router
+        ) {
```
eddiejaoude: I think this is more readable. Does the IDE not do this automatically?

```typescript
constructor(
   private authService: AuthenticationService,
   private projectService: ProjectService,
   private router: Router
) { }
```

web/src/app/projects/view/view.component.ts
webkhushboo: Done !!

web/src/resolvers/edit-project.resolver.ts
webkhushboo: Yes Id does it automatically , I have updated it initially. Will take Id formatting going forward

web/src/app/projects/view/view.component.ts
```diff
@@ -40,9 +40,9 @@ export class ViewProjectComponent implements OnInit {
     }
 
     ngOnInit(): void {
-        this.projectSubscription = this.projectService.findOneById(this.project.uid).subscribe((project: ProjectModel) => {
-                this.project = project;
-            });
+        this.projectSubscription = this.projectService
+            .findOneById(this.project.uid)
+            .subscribe((project: ProjectModel) => { this.project = project; });
```
eddiejaoude: the curly brackets are not needed 😄 

https://github.com/DashboardHub/PipelineDashboard/pull/1194#discussion_r279042504

web/src/resolvers/edit-project.resolver.ts
eddiejaoude: 👍

web/src/app/projects/view/view.component.ts
webkhushboo: Point to be noted !! Fixed

eddiejaoude: 👍 🤓

web/src/app/projects/edit/edit.component.ts
```diff
@@ -39,7 +42,9 @@ export class EditProjectComponent implements OnInit {
 
         this.projectSubscription = this.projectService
             .findOneById(this.uid)
-            .subscribe((project: ProjectModel) => this.projectForm.reset(project));
+            .subscribe((project: ProjectModel) => {
+                    this.projectForm.reset(project);
+            });
```
eddiejaoude: Same here, single line is possible

```typescript
.subscribe((project: ProjectModel) => this.projectForm.reset(project));
```

eddiejaoude:
> Chore(style):Removed unneccessary semicolon

Its best to keep all the commit message lowercase as you were doing before

PR #1196, opened 2019-04-29
Issue 1160

webkhushboo:
RELATES TO # 1160

### Notes

- [ ] Created main module and moved main component to main module
- [ ] Implemented lazy loading for main module in app-routing module
- [ ] Moved all the services to core module
- [ ] Move all guards, interceptors and resolvers to core module
- [ ] Moved pipes module under app folder
- [ ] Imported pipes module inside shared module
- [ ] Moved directives, dialog and models to the shared module
- [ ] Moved profile, features, privacy and terms and condition components under main module components folder
- [ ] Created projects module and added create , edit , view , repository components to this module
- [ ] Applied lazy loading on projects module and added routing in app-routinig
- [ ] Moved shared components like project and public project to shared module and reused them in main module as well in projects module
- [ ] Moved main component.html , main.component.scss and main.component.ts content to app.component.html , app.component.scss and app.component.ts to make app component as entry point for main module as well for projects module


webkhushboo:
Points to be fixed in PR
1. Fixed typos(spelling of refactor) in commit messages
2. Fixed issue no from 1192 to 1160 in commit message(#1192 moved directive to shared module)

web/src/app/projects/edit/edit.component.html
```diff
@@ -38,7 +38,7 @@
             <mat-divider></mat-divider>
             <mat-card-actions>
                 <div fxLayout="row" fxLayoutGap="20px" fxLayoutAlign="center center">
-                    <button mat-stroked-button type="button" [routerLink]="['/project', uid]">Cancel</button>
+                    <button mat-stroked-button type="button" (click)="onClickCancel()">Cancel</button>
```
eddiejaoude: Does this not add more code, rather than just having a `routerLink`?

eddiejaoude:
Looks good 👍 . 

1 small comment on moving from button `link` to `click`.

It might be worth adding to the `README.md` a few points for what each module is for, for example:

| module | description |
| :--- | :--- |
| core | ... |
| main | ... |
| shared | ... |
...

web/src/app/projects/edit/edit.component.html
webkhushboo: initially I was facing issue with relative link URL in routerLink so I have updated it click event. Now I found that we need to just go on its parent route when click on cancel so updated it again with routerLink ='../'

Thanks for pointing it out.

eddiejaoude: I am not sure about `../`, let me have a think 🤔

eddiejaoude: Can we not go from the root of the `uri` by prepending it with `/`?

What is the url its going `from` and `to`?

webkhushboo: It means its going one path up.
It is going from projects/:uid/edit to projects:uid .

eddiejaoude: Ok I will have a look, because I would have thought using `/projects/:uid` would work.

Lets merge it for now, the rest of it looks fine and I don't want to block you. I can do some research later on 🤓

webkhushboo: When I was on route "/projects/:uid/edit" and I click on cancel button using routerLink = ["projects", ''uid''] then it was appending this path in current route path which is wrong. Ideally we have to go one path up which can be achieved using '../'

Please let me know what do you think on this ?

webkhushboo:
@eddiejaoude 
Have you taken care of commit messages while merging PR ?

eddiejaoude:
@webkhushboo yep 👍 

I squashed the commits...

Screenshot 2019-05-01 11 25 34

PR #1203, opened 2019-05-04
chore(refactor): #1184 created one component for create and edit project

### LINKED ISSUE #1184: Create and edit should be created as reusable components under shared components
> Create and edit should be created as reusable components under shared components

From #1151

webkhushboo:
closes #1184

### Notes

- [ ] Created one component for create and edit project components
- [ ] Removed edit component

Kindly Review


web/src/app/projects/create-edit/create-edit.component.html
```diff
@@ -32,14 +32,16 @@
...
             </mat-card-content>
 
             <mat-divider></mat-divider>
             <mat-card-actions>
                 <div fxLayout="row" fxLayoutGap="20px" fxLayoutAlign="center center">
-                    <button mat-stroked-button type="button" routerLink="../">Cancel</button>
-                    <button mat-raised-button type="submit" color="accent" [disabled]="projectForm.invalid">Save</button>
+                    <button mat-stroked-button type="button" routerLink="../" color="warn">Cancel</button>
+                    <button mat-raised-button type="submit" color="accent" [disabled]="projectForm.invalid">
+                        <span *ngIf="!isEdit;else other">Create</span>
```
eddiejaoude: I think only this is needed in the condition?

```typescript
*ngIf="!isEdit"
```

web/src/app/projects/create-edit/create-edit.component.ts
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    // This function will create project and edit project details based upon if click on edit or add
+    create(): void {
+        if (this.uid) {
+            this.createEditSubscription = this.projectService
+                .save({ uid: this.uid, ...this.projectForm.getRawValue() })
+                .pipe(
+                    catchError((error: any): any => this.snackBar.open(error.message, undefined, { duration: 5000 }))
```
eddiejaoude: Is it possible to not use `any`? This can be done later as a separate ticket to update the whole application to remove `any` and use the correct types

web/src/app/projects/create-edit/create-edit.component.html
```diff
@@ -1,8 +1,8 @@
 <div class="mat-content" fxLayout="column" fxLayoutGap="20px">
-    <form (submit)="save()" [formGroup]="projectForm">
+    <form (submit)="create()" [formGroup]="projectForm">
```
eddiejaoude: Because the form is for `create` and `edit`, I think the naming `save` would be more appropriate

eddiejaoude:
Looks good 👍 , a few small comments

web/src/app/projects/create-edit/create-edit.component.html
webkhushboo: Good catch!! Removed 👍

web/src/app/projects/create-edit/create-edit.component.html
webkhushboo: Review incorporated. 👍

web/src/app/projects/create-edit/create-edit.component.ts
webkhushboo: Here snackbar is showing error message when error occurred so its returning MatSnackBarRef<SimpleSnackBar> instance but in this catchError function we are not returning anything. 
Also when I am trying to add return type as void for this arrow function its not allowing in catchError argument type. 
Let me know if you have suggest any better option for it ?

eddiejaoude: Ok, it's something we can look at later 🤓

Issue #1204, opened 2019-05-04
Spinner when loading data


eddiejaoude:
> If there is delay in data fetching and response apply on UI then there should be some loader or spinner displayed to user so that he knows something is happening in background. For e.g is user loads the home page and public projects are displayed in few seconds then there should be some spinner

From #1153

PR #1218, opened 2019-05-06
chore(bug): #1156 fixed issue user info not shown until after an action

### LINKED ISSUE #1156: After login, user info not shown until after an action
Found in #1153

Once an action is performed on the page, it shows the user has logged in.

webkhushboo:
closes #1156

- [ ] UI: Added output event emitter to refresh home route
- [ ] UI: As soon as user logged in it will navigate to home page

Kindly review

### DIFF
--- a/web/src/app/app.component.ts
+++ b/web/src/app/app.component.ts
@@ -1,4 +1,5 @@
-import { Component } from '@angular/core';
+import { Component, OnInit } from '@angular/core';
+import { Router } from '@angular/router';
 
 // Dashboard hub Icon register
 import { DomSanitizer } from '@angular/platform-browser';
@@ -16,7 +17,7 @@ import { environment } from './../environments/environment';
     templateUrl: './app.component.html',
     styleUrls: ['./app.component.scss']
 })
-export class AppComponent {
+export class AppComponent implements OnInit {
     public version: string;
     public publicRoutes: Navigation[] = [
         {
@@ -68,6 +69,7 @@ export class AppComponent {
         private _iconRegistry: MatIconRegistry,
         private _domSanitizer: DomSanitizer,
         private authService: AuthenticationService,
+        private router: Router
     ) {
         this._iconRegistry
             .addSvgIconInNamespace('assets', 'dashboardhub',
@@ -110,5 +112,13 @@ export class AppComponent {
 
     public showDoorbell(): void {
         (<any>window).doorbell.show();
+    }
+
+    ngOnInit(): void {
+        this.authService.getLoginEmitter().subscribe((result: boolean) => {
+            if (result) {
+                this.router.navigate(['/']);
+            }
+        });
     }
 }
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,4 @@
-RELATES TO # {issue_number}
+closes #{issue_number}
 
 ### Notes
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,4 +29,4 @@ deploy:
       secure: $FIREBASE_TEST_TOKEN
     project: "pipelinedashboard-test"
     on:
-      branch: firebase-mvp
+      branch: v0.11.5
--- a/web/src/app/core/services/authentication.service.ts
+++ b/web/src/app/core/services/authentication.service.ts
@@ -1,5 +1,5 @@
 import { Injectable, Output, EventEmitter } from '@angular/core';
-
+import { Router } from '@angular/router';
 // Firestore modules
 import { AngularFireAuth } from '@angular/fire/auth';
 import { AngularFirestore } from '@angular/fire/firestore';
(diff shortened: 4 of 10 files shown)


eddiejaoude:
Error when logging out (after redirect to the homepage)...

Screenshot 2019-05-07 at 06 19 33

web/src/app/app.component.ts
```diff
@@ -111,4 +113,12 @@ export class AppComponent {
     public showDoorbell(): void {
         (<any>window).doorbell.show();
     }
+
+    ngOnInit(): void {
+        this.authService.getLoginEmitter().subscribe((result: boolean) => {
+            if (result) {
+                this.router.navigate(['/']);
```
eddiejaoude: I don't think it is needed to redirect the user to the homepage on login. 

Also I think an event emitter might be too much, this file has access to the `isAuthenticated` variable from the `AuthService` https://github.com/DashboardHub/PipelineDashboard/pull/1218/files#diff-9fb615cefc2705778e04e2ef31bec93bR24

It is difficult for me to test, because the issue of not having the user details appear until they perform an action seems random

eddiejaoude:
I created an alternative from your branch, trying to simplify the event emitter. It seems to work but the issue is difficult for me to replicate so I can not test it properly. @webkhushboo let me know what you think?

web/src/app/app.component.ts
webkhushboo: What should happen after login ? Where should it go instead homepage ?

webkhushboo:
@eddiejaoude 
I have added event emitter to let the app component know there is some change in route after login and it should navigate to home.

webkhushboo:
What alternative you have tried ? Can you please explain ?

webkhushboo:
It takes a few second to show the logout icon after login in home page. We need to add some kind of loader to show the activity for fetching the data .

web/src/app/app.component.ts
eddiejaoude: Can the user not stay on the page they are currently on?

For example, if they are view a dashboard, then they decide to log in, they should stay on the same dashboard

eddiejaoude:
> It takes a few second to show the logout icon after login in home page. We need to add some kind of loader to show the activity for fetching the data .

Yes, I think a simple loader might be better, do you have any examples you have seen on the Internet, I will take a look too 🤓

webkhushboo:
Please see the below two examples of spinner.
1. ngx-spinner 
2. Angular material spinner

https://napster2210.github.io/ngx-spinner/
https://material.angular.io/components/progress-spinner/overview

Please let me know which one you prefer?

eddiejaoude:
The material spinner is better. Where would you place it? We don't want to block the whole page, but just to notify the user that something is happening.

What about using something like this, where every http request shows a progress, similar to GitHub  https://github.com/murhafsousli/ngx-progressbar

webkhushboo:
We have similar progress bar in material too . Please see the below link
https://material.angular.io/components/progress-bar/overview

webkhushboo:
We should place the spinner in the root level and hide it once the response complete using subject and observable or what do you suggest?

webkhushboo:
We also have one issue for showing spinner https://github.com/DashboardHub/PipelineDashboard/issues/1204

eddiejaoude:
> We also have one issue for showing spinner #1204

Well remembered! We should do the progress work under #1204 

> We should place the spinner in the root level and hide it once the response complete using subject and observable or what do you suggest?

We should be able to use an interceptor to effect all http requests https://angular.io/guide/http#intercepting-requests-and-responses

webkhushboo:
But we use interceptors to modify the http request , Here we have to show the spinner as soon as we send request from our component and hide it once the response come in subscribe method.

webkhushboo:
image

Something like this ?

eddiejaoude:
Can it go right at the top of the page, at the top of the orange, with an absolute position on top so it doesn't push everything down when it appears

eddiejaoude:
This PR has some conflicts.

I think it would be good to keep the logout redirect to the home page, but the login probably is not needed, the user can stay on the page they are currently on when logging in

webkhushboo:
@eddiejaoude 
Should I remove redirecting to home after login ?

webkhushboo:
@eddiejaoude 
Do you want to change the position of spinner to the top ?

eddiejaoude:
Yes, I think so. Remove the redirect after login.

The spinner in another PR is higher, directly under the orange navbar, is this not the same spinner?

webkhushboo:
Yes that image is older . Please refer this image . 

image

webkhushboo:
@eddiejaoude 
I will remove the redirect after login

eddiejaoude:
Looks good. Thank you 👍

There is also 1 conflict to resolve after merging another PR

webkhushboo:
@eddiejaoude 

As we don't need to redirect to the home page after login and this PR contains most of those changes.
Can we discard this PR ? Also this PR has not latest changes to merge conflicts are also there .
I will fix the logout issue in another PR as it is a very small change.

webkhushboo:
Closing this pull request as we don't need login redirect changes to homepage.
And raised another PR for logout redirect changes
https://github.com/DashboardHub/PipelineDashboard/pull/1226

eddiejaoude:
Thank you for putting a comment and link to the other relevant PR 👍 🤓

PR #1219, opened 2019-05-07
 feat(auth): #1156 removed event emitter for auth

eddiejaoude:
As an alternative to PR #1218


webkhushboo:
Closing this pull request as we don't need to redirect on home after login.

PR #1222, opened 2019-05-08
feat(project): #1204 added progress bar in the app

### LINKED ISSUE #1204: Spinner when loading data

webkhushboo:
closes #1204

### Notes

- [ ] UI: Added spinner in the app

Note : 
1. Right now I have added the subject and observable for setting progress bar status in authentication service. Do you want me to create new service for it or does it look good?
2. I have added the spinner for login and public project component . Please let me know if it looks okay or do you want to some kind of change?
3. For firebase requests we can not make change in http interceptors. We have to show hide the spinner based upon the response for each component .  Although I have added spinner to main component and created subject and observable to track the status of it 

Please have a look on it and share your feedback on same.


eddiejaoude:
Can you add a screenshot please, thank you

web/src/app/core/services/authentication.service.ts
```diff
@@ -21,6 +21,7 @@ import { ProfileModel, LoginAuditModel } from '../../shared/models/index.model';
 })
 export class AuthenticationService {
 
+    private spinnerSubject: Subject<Boolean> = new Subject();
```
eddiejaoude: This would be better in its own service, so the `AuthenticationService` is not included in many places where the `AuthenticationService` is not really required

web/src/app/core/services/authentication.service.ts
```diff
@@ -29,7 +30,7 @@ export class AuthenticationService {
         private afs: AngularFirestore,
         private fns: AngularFireFunctions,
         private deviceService: DeviceDetectorService,
-    ) {
+        ) {
```
eddiejaoude: The extra indentation is not required

web/src/app/core/services/authentication.service.ts
```diff
@@ -144,6 +149,16 @@ export class AuthenticationService {
             });
     }
 
+    // This function will set the progress bar status
+    public setProgressBar(status: boolean): void {
+       this.spinnerSubject.next(status);
+    }
+
+    // this function will return the status of progress bar to main component
+    public getProgressBar(): Observable<Boolean> {
```
eddiejaoude: Both these functions to be moved to its own service as mentioned here https://github.com/DashboardHub/PipelineDashboard/pull/1222/files#r282212369

web/src/app/main/main.component.scss
```diff
@@ -0,0 +1,3 @@
+.margin-top-10{
+    margin-top: 10px;
+}
```
eddiejaoude: File requires a blank line ending

web/src/app/main/main.component.ts
```diff
@@ -1,10 +1,23 @@
...
+export class MainComponent implements OnInit {
 
+    public showSpinner: boolean = true;
+
+    constructor(
+        private authService: AuthenticationService
+    ) { }
+
+    ngOnInit(): void {
+        this.authService.getProgressBar().subscribe((data: boolean) => {
```
eddiejaoude: Easier to read when on multiple links

```typescript
this.authService
   .getProgressBar()
   .subscribe((data: boolean) => {

   });
```

eddiejaoude:
I only had a quick look but looks good 👍, I left a few comments 🤓

webkhushboo:
> please

image

web/src/app/core/services/authentication.service.ts
webkhushboo: Created spinner service for the function and removed from authentication service. Thanks for suggestion

web/src/app/core/services/authentication.service.ts
webkhushboo: Removed the extra indentation

web/src/app/main/main.component.scss
webkhushboo: Done

web/src/app/main/main.component.ts
webkhushboo: Done

eddiejaoude:
Not sure I understand the screenshot above?

eddiejaoude:
I checked out the branch and now I see it in the screenshot above, the blue line 👍 . It looks great! We should apply it to move places, but that can be another issue 🤓

Issue #1224, opened 2019-05-09
User logging in progress bar should be available on all pages


eddiejaoude:
I noticed when logging in from a dashboard page the progress bar is not shown.

webkhushboo:
![image](https://user-images.githubusercontent.com/10898595/57545094-ea8daf80-7376-11e9-83a0-143341757a14.png)


I can see the progress bar on dashboard page . Can you please check again
Please see the attached screenshot for your reference.

eddiejaoude:
The screenshot is the home page. If you click on one of the dashboards, then log in?

webkhushboo:
One of the dashboard means, if I click on project ? Is  that you are referring to or something else ?
I agree this will be shown for all components under main module like Home page , help , privacy , profile 
and will not be shown for projects module components. 
I will make those changes now .

Right now I have added the spinner only in main module. Now I will add it to project module too.
Please correct me If I am understanding something wrong.

eddiejaoude:
Yes sorry, by dashboard, I mean the project page (project dashboard) 👍

It needs to be available from all pages because people could login from any page

PR #1225, opened 2019-05-10
defect(project): #1224 added spinner to app module

webkhushboo:
closes # 1224

### Notes

- [ ] UI: Added spinner to app module
- [ ] Spinner should be visible to all pages when user login

Kindly review

### DIFF
--- a/web/src/app/shared/components/public/public.component.ts
+++ b/web/src/app/shared/components/public/public.component.ts
@@ -25,20 +25,17 @@ export class PublicProjectsComponent implements OnInit {
     }
 
     ngOnInit(): void {
-        this.spinnerService.setProgressBar(true);
         if (this.router.url === '/') {
             this.projectSubscription = this.projectService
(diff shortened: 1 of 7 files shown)


web/src/app/shared/components/public/public.component.ts
```diff
@@ -25,20 +25,17 @@ export class PublicProjectsComponent implements OnInit {
     }
 
     ngOnInit(): void {
-        this.spinnerService.setProgressBar(true);
```
eddiejaoude: Will it now be triggered by the `service`?

eddiejaoude:
Looks good 👍 , one small question on when it will be used

web/src/app/shared/components/public/public.component.ts
eddiejaoude: Ignore me, that was part of another PR 🤦‍♂ . Here we are just moving the location of it.

PR #1230, opened 2019-05-12
 chore(refactor): #1229 move spinner to taps

### LINKED ISSUE #1229: Move spinner triggers to taps

eddiejaoude:
closes #1229 

### Notes

- [ ] moved spinner to pre/post `tap` - looks the same, starts slightly later and finishes the same time
- [ ] removed unused spinner imports

### DIFF
--- a/web/src/app/core/services/authentication.service.ts
+++ b/web/src/app/core/services/authentication.service.ts
@@ -9,7 +9,7 @@ import { auth, User } from 'firebase/app';
 
 // Rxjs operators
 import { from, Observable, of } from 'rxjs';
-import { filter, concatMap, switchMap, first, takeUntil } from 'rxjs/operators';
+import { filter, concatMap, switchMap, first, takeUntil, tap } from 'rxjs/operators';
 
 // Third party modules
 import { DeviceDetectorService } from 'ngx-device-detector';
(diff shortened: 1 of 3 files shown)


eddiejaoude:
@webkhushboo please review when you get time 🤓

web/src/app/core/services/authentication.service.ts
```diff
@@ -84,11 +84,11 @@ export class AuthenticationService {
 
     // This function used to login via github
     public login(): void {
-        this.spinnerService.setProgressBar(true);
         const provider: auth.GithubAuthProvider = new auth.GithubAuthProvider();
         provider.addScope('repo,admin:repo_hook');
         from(this.afAuth.auth.signInWithPopup(provider))
             .pipe(
+                tap(() => this.spinnerService.setProgressBar(true)),
```
webkhushboo: Initially we were starting spinner , when we click on login button now after adding tap , its starting spinner once you enter sign in details in popup . Do you want this change intentionally ?

eddiejaoude: My concern is, if someone closes the small popup window without signing in, the progress spinner keeps going and never ends (plus I think it is a bit cleaner code to have it in the `pipe`). So I think it makes more sense for the progress bar to start once the small popup closes after the user has signed in.

webkhushboo: Thanks for clarification . Looks good 👍

PR #1238, opened 2019-05-14
feature(project): #1227 added index file for models and pipes

### LINKED ISSUE #1227: Services, pipes, models to use index file

webkhushboo:
closes #1227 

### Notes

- [ ] Updated index.model file for navigation and user model
- [ ] Added index file for pipes and updated its referenced

### DIFF
--- a/web/src/app/core/services/index.service.ts
+++ b/web/src/app/core/services/index.service.ts
@@ -1,0 +1,5 @@
+export * from './authentication.service';
+export * from './project.service';
+export * from './repository.service';
+export * from './user.service';
+export * from './spinner.service';
--- a/web/src/app/core/services/index.service.ts
+++ b/web/src/app/core/services/index.service.ts
@@ -1,5 +1,5 @@
 export * from './authentication.service';
 export * from './project.service';
 export * from './repository.service';
-export * from './user.service';
 export * from './spinner.service';
+export * from './user.service';
--- a/web/src/app/core/services/repository.service.ts
+++ b/web/src/app/core/services/repository.service.ts
@@ -10,8 +10,7 @@ import { pluck } from 'rxjs/operators';
 
 // Dashboard hub model and services
 import { AuthenticationService } from './authentication.service';
-import { RepositoriesModel } from '../../shared/models/index.model';
-import { RepositoryModel, ProfileModel } from '../../shared/models/index.model';
+import { RepositoryModel, RepositoriesModel, ProfileModel } from '../../shared/models/index.model';
 
 @Injectable({
     providedIn: 'root'
--- a/web/src/app/shared/models/activity.model.ts
+++ b/web/src/app/shared/models/activity.model.ts
@@ -1,4 +1,4 @@
-import { RepositoryModel } from './repository.model';
+import { RepositoryModel } from './index.model';
 
 export class ActivityModel {
     type: string;
--- a/web/src/app/shared/models/activity.model.ts
+++ b/web/src/app/shared/models/activity.model.ts
@@ -1,4 +1,4 @@
-import { RepositoryModel } from './index.model';
+import { RepositoryModel } from './repository.model';
 
 export class ActivityModel {
     type: string;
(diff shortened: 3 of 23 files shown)


webkhushboo:
@eddiejaoude 
General question: Why do we need index file for services? If I am adding index file then importing the services using index.service file . 
Its causing circular dependency issues
repostory.service.ts -- > index.service.ts --> repository.service.ts file 

Please let me know your reviews on it

eddiejaoude:
The benefit of the index file, is that only a single import is required, for example

```typescript
import { DigitsPipe, UserPipe, EnvironmentTypePipe } from './index.pipe';
```

eddiejaoude:
Should not get a circular reference, it is technically the same.

It goes from...
```
service -> modelA
service -> modelB
service -> modelC
```

to

```
service -> model index -> modelA / modelB / modelC
```

eddiejaoude:
> Its causing circular dependency issues

Yes, sorry, just realised, anything in the same directory does not use `index`. For example, `serviceA` would import `serviceB` but a component would use the `index`

webkhushboo:
@eddiejaoude 
Do we really need index file for services ?

eddiejaoude:
Yep, the same again. Any file in service directory does not use the `index.services`, but everywhere else does. Then you won't get the circular reference

webkhushboo:
You mean in repository.service file we will be importing authentication.service.ts directly ?

eddiejaoude:
Yep 👍 . Any directory that has an `index`, those file don't use the `index` themselves.

webkhushboo:
great 👍

web/src/app/shared/models/profile.model.ts
```diff
@@ -1,4 +1,4 @@
-import { ActivityModel } from './activity.model';
+import { ActivityModel } from './index.model';
```
eddiejaoude: The models themselves import the files directly. This change is not required

web/src/app/shared/models/event.model.ts
```diff
@@ -1,5 +1,4 @@
-import { UserModel } from './user.model';
-import { RepositoryModel } from './repository.model';
+import { RepositoryModel, UserModel } from './index.model';
```
eddiejaoude: These 2 original imports are correct

web/src/app/shared/models/activity.model.ts
```diff
@@ -1,4 +1,4 @@
-import { RepositoryModel } from './repository.model';
+import { RepositoryModel } from './index.model';
```
eddiejaoude: Same here, original import is fine as its loading files in the same type/directory (in this case `model`s)

web/src/app/core/services/repository.service.ts
```diff
@@ -10,8 +10,7 @@ import { pluck } from 'rxjs/operators';
 
 // Dashboard hub model and services
 import { AuthenticationService } from './authentication.service';
-import { RepositoriesModel } from '../../shared/models/index.model';
-import { RepositoryModel, ProfileModel } from '../../shared/models/index.model';
+import { RepositoryModel, RepositoriesModel, ProfileModel } from '../../shared/models/index.model';
```
eddiejaoude: Looks good 👍

eddiejaoude:
Looks good 👍 , a few comments where I think `index` should not be used as its in the same directory (type), for example model importing another model should be direct and not via the `index`

web/src/app/core/services/index.service.ts
```diff
@@ -0,0 +1,5 @@
+export * from './authentication.service';
+export * from './project.service';
+export * from './repository.service';
+export * from './user.service';
+export * from './spinner.service';
```
eddiejaoude: `spinner` should be before `user`

webkhushboo: done

web/src/app/core/services/repository.service.ts
webkhushboo: Thanks

web/src/app/shared/models/activity.model.ts
webkhushboo: done

web/src/app/shared/models/event.model.ts
webkhushboo: done

web/src/app/shared/models/profile.model.ts
webkhushboo: done

PR #1253, opened 2019-05-21
defect(project): #1240 updated new date to timestamp object to fix da…

### LINKED ISSUE #1240: Fix Firestore date warnings
The dates stored in `Firestore` need updating to be their new format. Current style is deprecated.

**Note: I might need to delete the data in the current Firestore **

Error from the logs

```
The behavior for Date objects stored in Firestore is going to change
AND YOUR APP MAY BREAK.
To hide this warning and ensure your app does not break, you need to add the
following code to your app before calling any other Cloud Firestore methods:

  const firestore = new Firestore();
  const settings = {/* your settings... */ timestampsInSnapshots: true};
  firestore.settings(settings);

With this change, timestamps stored in Cloud Firestore will be read back as
Firebase Timestamp objects instead of as system Date objects. So you will also
need to update code expecting a Date to instead expect a Timestamp. For example:

  // Old:
  const date = snapshot.get('created_at');
  // New:
  const timestamp = snapshot.get('created_at');
  const date = timestamp.toDate();

Please audit all existing usages of Date when you enable the new behavior. In a
future release, the behavior will change to the new behavior, so if you do not
follow these steps, YOUR APP MAY BREAK.
```

webkhushboo:
closes #1240 

### Notes

- [ ] Fixed new Date() to timestamp object to fix the date warnings

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,3 +102,18 @@ The format is `<type>(<scope>): #<issue-no> <subject>`, examples...
 | shared | Contains all shared components, directives, dialogs and models |
 | projects | Contains all components like create, edit, private , view projects and repository |
 | pipes | Contains all custom pipes |
+
+
+## To set up admin credentials for emulated functions:
+
+1. Open the Service Accounts pane of the Google Cloud Console.
+2. Make sure that App Engine default service account is selected, and use the options menu at right to select Create key.
+3. When prompted, select JSON for the key type, and click Create.
+4. Set your Google default credentials to point to the downloaded key:
+
+## Deploy firebase functions on local
+
+1. `set GOOGLE_APPLICATION_CREDENTIALS=path\to\key.json`
+2. `firebase setup:emulators:firestore`
+3. `firebase emulators:start`
+
--- a/web/src/app/core/services/authentication.service.ts
+++ b/web/src/app/core/services/authentication.service.ts
@@ -5,6 +5,7 @@ import { Router } from '@angular/router';
 import { AngularFireAuth } from '@angular/fire/auth';
 import { AngularFirestore } from '@angular/fire/firestore';
 import { AngularFireFunctions } from '@angular/fire/functions';
+import * as firebase from 'firebase';
 import { auth, User } from 'firebase/app';
 
 // Rxjs operators
@@ -15,7 +16,6 @@ import { concatMap, filter, first, switchMap, takeUntil, tap } from 'rxjs/operat
 import { DeviceDetectorService } from 'ngx-device-detector';
 
 // Dashboard hub models
-import * as firebase from 'firebase';
 import { LoginAuditModel, ProfileModel } from '../../shared/models/index.model';
 import { SpinnerService } from './spinner.service';
 
--- a/web/src/app/core/services/authentication.service.ts
+++ b/web/src/app/core/services/authentication.service.ts
@@ -15,6 +15,7 @@ import { concatMap, filter, first, switchMap, takeUntil, tap } from 'rxjs/operat
 import { DeviceDetectorService } from 'ngx-device-detector';
 
 // Dashboard hub models
+import * as firebase from 'firebase';
 import { LoginAuditModel, ProfileModel } from '../../shared/models/index.model';
 import { SpinnerService } from './spinner.service';
 
@@ -125,7 +126,7 @@ export class AuthenticationService {
             .doc<ProfileModel>(profile.uid)
             .collection<LoginAuditModel>('logins')
             .add({
-              date: new Date(),
+              date: firebase.firestore.Timestamp.fromDate(new Date()),
               userAgent: this.deviceService.getDeviceInfo().userAgent,
               os: this.deviceService.getDeviceInfo().os,
               browser: this.deviceService.getDeviceInfo().browser,
--- a/web/src/app/core/services/project.service.ts
+++ b/web/src/app/core/services/project.service.ts
@@ -1,10 +1,10 @@
 import { Injectable } from '@angular/core';
 import { AngularFirestore } from '@angular/fire/firestore';
+import * as firebase from 'firebase';
 import { from, of, Observable } from 'rxjs';
 import { v4 as uuid } from 'uuid';
 
 // Dashboard model and services
-import * as firebase from 'firebase';
 import { ProjectModel, RepositoryModel } from '../../shared/models/index.model';
 import { AuthenticationService } from './authentication.service';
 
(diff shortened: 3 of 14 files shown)


webkhushboo:
@eddiejaoude 

I have updated the new Date() to firebase.firestore.Timestamp.fromDate(new Date()) in services.

Kindly review.

web/src/app/core/services/authentication.service.ts
```diff
@@ -15,6 +15,7 @@ import { concatMap, filter, first, switchMap, takeUntil, tap } from 'rxjs/operat
 import { DeviceDetectorService } from 'ngx-device-detector';
 
 // Dashboard hub models
+import * as firebase from 'firebase';
```
eddiejaoude: This should probably go in the import group above

web/src/app/core/services/project.service.ts
```diff
@@ -25,8 +26,8 @@ export class ProjectService {
       access: { admin: [this.authService.profile.uid] },
       ...data,
       repositories: [],
-      createdOn: new Date(),
-      updatedOn: new Date(),
+      createdOn: firebase.firestore.Timestamp.fromDate(new Date()),
+      updatedOn: firebase.firestore.Timestamp.fromDate(new Date()),
```
eddiejaoude: This might be easier to read if `declared` at the top under the imports?

```typescript
declare type Timestamp = firebase.firestore.Timestamp;
```

Then used with:


```typescript
createdOn: Timestamp.fromDate(new Date()),
updatedOn: Timestamp.fromDate(new Date()),
```

web/src/app/shared/models/project.model.ts
```diff
@@ -1,3 +1,4 @@
+import { firestore } from 'firebase';
```
eddiejaoude: Should these be group separately, the same as other places?

eddiejaoude:
Also what about the `functions`?

web/src/app/core/services/project.service.ts
webkhushboo: @eddiejaoude 
 I have tried the same way as you suggested. But its giving me error "Timestamp is declared as type but is used as a value". Please see the attached screenshot for your reference.

image

web/src/app/shared/models/project.model.ts
webkhushboo: Done

web/src/app/core/services/authentication.service.ts
webkhushboo: Done

web/src/app/core/services/project.service.ts
eddiejaoude: Ah ok, yes it won't work, I was thinking of types not value 🤦‍♂ . Thank you for trying 👍

webkhushboo:
We need firebase dependency for this which was not present in functions. Should I add it ?

eddiejaoude:
> We need firebase dependency for this which was not present in functions. Should I add it ?

Good question, I think updating `date` to the new style is important, and if a another dependency is needed, then it should be fine 👍

webkhushboo:
Thanks for confirming. I have added the dependency , can you please confirm if I have to deploy the functions to verify them ?

eddiejaoude:
Yes, I think it is best to double check when deployed if possible

webkhushboo:
@eddiejaoude 
Just for updating you , we can not use firebase dependency in the functions .
Getting this error 
“Cannot find name 'ServiceWorkerRegistration'” error when creating a firebase cloud function using Typescript

Although I found the solution for it. We should be using firebase-admin here instead firebase.

eddiejaoude:
@webkhushboo yes you are right 👍 , I did a quick search and saw this https://firebase.google.com/docs/reference/js/firebase.database.ServerValue

README.md
```diff
@@ -102,3 +102,18 @@ The format is `<type>(<scope>): #<issue-no> <subject>`, examples...
...
+1. Open the Service Accounts pane of the Google Cloud Console.
+2. Make sure that App Engine default service account is selected, and use the options menu at right to select Create key.
+3. When prompted, select JSON for the key type, and click Create.
+4. Set your Google default credentials to point to the downloaded key:
+
+## Deploy firebase functions on local
+
+1. `set GOOGLE_APPLICATION_CREDENTIALS=path\to\key.json`
+2. `firebase setup:emulators:firestore`
+3. `firebase emulators:start`
```
eddiejaoude: Is there anything to change on the UI to know to look locally rather than on Firebase Cloud?

webkhushboo:
@eddiejaoude 
I have updated the node version to 10 in package.json for functions . that's the reason its showing the check failed. 
Can you please update the node version ?

eddiejaoude:
@webkhushboo you can update it in `.travis.yml` file, change line `3` for the `nodejs` version. But if you are away from your computer I can do it?

eddiejaoude:
I updated it 🤓

eddiejaoude:
Oh I will need to update the cloud functions also

eddiejaoude:
You did it already, ok great 👍 . CI is building now 🤓

eddiejaoude:
CI still broken...

```
3.67s$ (cd functions; npm run build)
> functions@ build /home/travis/build/DashboardHub/PipelineDashboard/functions
> tsc
node_modules/gaxios/build/src/index.d.ts(14,66): error TS1005: '>' expected.
node_modules/gaxios/build/src/index.d.ts(14,103): error TS1109: Expression expected.
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! functions@ build: `tsc`
npm ERR! Exit status 2
npm ERR! 
npm ERR! Failed at the functions@ build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm ERR! A complete log of this run can be found in:
npm ERR!     /home/travis/.npm/_logs/2019-05-22T19_56_20_578Z-debug.log
The command "(cd functions; npm run build)" exited with 2.
```

I get it locally also with 

```
node -v
v10.15.2
```

I found [this](https://github.com/firebase/firebase-functions/issues/404) and I tried updating all the function dependencies to the latest but the error still remained.

webkhushboo:
image


After upgrading typescript to latest version . It seems to be working on my local. Please see the attached screenshot for your reference.

eddiejaoude:
Ok interesting. Can you push it up and we will let CI decided 🤓 - if CI is happy, then I am happy 😎

eddiejaoude:
Looks good 👍

eddiejaoude:
I updated the other out of date dependencies also...

```
functions git:(issue-1240) npm outdated
Package             Current  Wanted  Latest  Location
firebase-functions    2.1.0   2.3.1   2.3.1  functions
tslint                5.8.0   5.8.0  5.16.0  functions
```

PR #1323, opened 2019-06-19
fix(repo): #1322 when exception is thrown dont write to db

### LINKED ISSUE #1322: Repository info write error, when exception is throw
`functions/src/repository/info.ts`

```
Unhandled error Error: Value for argument "data" is not a valid Firestore document. Input is not a plain JavaScript object
```

eddiejaoude:
closes #1322

### Notes

- [x] functions: if an exception is thrown, we should not try to write to the DB


webkhushboo:
How can i check which error is fixed in this PR ?

eddiejaoude:
> How can i check which error is fixed in this PR ?

Good question 👍 ! It is very difficult, we would need an exception to be thrown - its an edge case. You might have just do it by eye

PR #1332, opened 2019-06-21
WIP:feat(tokens): #1301 CRUD tokens for project

kykyk:
WIP:#1301 CRUD tokens for project

### Notes
bug: permission denied for create

Screenshot from 2019-06-21 11-04-04
Screenshot from 2019-06-21 11-01-54
Screenshot from 2019-06-21 10-58-32
Screenshot from 2019-06-21 10-58-12

PR #1338, opened 2019-06-27
feat(monitors): #1299 CRUD operation for monitor

### LINKED ISSUE #1299: CRUD monitors for project
Within a project we need to add the CRUD **monitors** functionality for `admins` (see screenshot below).

Screenshot 2019-06-12 11 51 17

## Steps

1. Click new `monitors` button in a project
2. In the new monitors section, there will be `list` page with `delete`, `create` and `edit` page

## Validation

- name is required must be unique within project and have more than 3 characters
- path must be unique with the project

```typescript
// monitor will be under project model
interface Monitor {
   uid: string;
   name: string;
   method:  'GET' | 'POST';
   path: string;
   expectedCode: string;
   expectedText?: string; 
   expectedDuration: number;
}
```

*Notes: Keep the styling vanilla as possible, this will come later*

webkhushboo:
closes #1299 

### Notes

- [ ] Add new monitor
- [ ] Edit monitor
- [ ] Delete monitor
- [ ] Show monitor list


web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
```
eddiejaoude: We don't haver spec files. We will introduce Cypress in the near future but not for the moment

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
```diff
@@ -0,0 +1,108 @@
...
+   */
+  ngDestroy(): void {
+    this.createEditSubscription.unsubscribe();
+    this.projectSubscription.unsubscribe();
+  }
+
+  /**
+   * This function is used to add the monitor in the array
+   */
+  addMonitorToList(): void {
```
eddiejaoude: I think `addMonitor` is descriptive enough for the name

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
```diff
@@ -0,0 +1,108 @@
...
+      name: [undefined, [Validators.required, Validators.minLength(3)]],
+      method: [undefined, Validators.required],
+      path: [undefined, Validators.required],
+      expectedCode: [undefined, Validators.required],
+      expectedText: [undefined],
+    });
+    this.projectSubscription = this.projectService
+      .findOneById(this.uid)
+      .subscribe((project: ProjectModel) => {
+        this.project = project;
```
eddiejaoude: I don't think this is needed any where else, so could be a local variable rathe than a class variable

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
```diff
@@ -0,0 +1,108 @@
...
+        this.monitorsList = this.project.monitors ? this.project.monitors : [];
+        this.id = this.route.snapshot.paramMap.get('id');
+        if (this.id) {
+          this.isEdit = true;
+          const filteredMonitor: MonitorModel[] = this.monitorsList.filter((monitor: MonitorModel) => {
```
eddiejaoude: This component could be simplified if we move the monitor logic to a monitor service

web/src/app/core/services/project.service.ts
```diff
@@ -147,4 +147,37 @@ export class ProjectService {
             { merge: true }))
       );
   }
+
+   // This function add the repository in any project
+   public saveMonitors(uid: string, monitors: MonitorModel[]): Observable<void> {
```
eddiejaoude: I think this could go in its own service.

I probably need to move adding a repository to its own service too. I will create a ticket for the repository refactor

web/src/app/monitors/monitors-list/monitors-list.component.html
```diff
@@ -0,0 +1,18 @@
+<mat-card>
+  <div fxLayout="row">
+    <div fxFlex="80%">
+      <h3>Monitors</h3>
+    </div>
+    <div fxFlex="20%"><button mat-raised-button type="submit" color="accent" routerLink="create">Add new monitor</button></div>
```
eddiejaoude: `Add monitor` is probably enough

web/src/app/monitors/monitors-list/monitors-list.component.html
```diff
@@ -0,0 +1,18 @@
...
+  </div>
+  <mat-list>
+    <mat-list-item fxLayout="column" *ngFor="let monitor of monitors">
+      <h4 mat-line>{{monitor.name}}</h4>
+      <p mat-line>{{monitor.method}} {{monitor.path}} {{monitor.expectedCode}} {{monitor.expectedText}}</p>
+      <p class="text-right">
+        <button mat-raised-button color="accent" (click)="editMonitor(monitor.uid)">Edit</button>
```
eddiejaoude: I think this would be better as a route link

web/src/app/projects/view/view.component.html
```diff
@@ -6,6 +6,7 @@
     <span>{{ project.title }}</span>
     <span fxFlex></span>
     <span fxLayout="row" fxLayoutGap="20px" fxLayoutAlign="center center" *ngIf="isAdmin()">
+      <button mat-raised-button type="submit" color="accent" routerLink="monitors">Monitors</button>
       <button mat-raised-button type="submit" color="accent" (click)="addRepository()">Add Repository</button>
```
eddiejaoude: I know this is not part of your PR, but when you make some edits for your PR, could you please rename `Add Repository` to just `Repositories`, thank you 🤓

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
```diff
@@ -0,0 +1,108 @@
...
+  ngDestroy(): void {
+    this.createEditSubscription.unsubscribe();
+    this.projectSubscription.unsubscribe();
+  }
+
+  /**
+   * This function is used to add the monitor in the array
+   */
+  addMonitorToList(): void {
+    this.monitorsList.push(this.monitorForm.value);
```
eddiejaoude: The monitor is missing a `uid`, I think thats why edit a monitor is not working

web/src/app/monitors/monitors-list/monitors-list.component.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * This method is used to delete the monitor from list
+   */
+  deleteMonitor(id: string): void {
+    const objIndex: number = this.monitors.findIndex((monitor: MonitorModel) => monitor.uid === id);
+    this.monitors = [
+      ...this.monitors.slice(0, objIndex),
+      ...this.monitors.slice(objIndex + 1),
```
eddiejaoude: I think this would be cleaner with a `filter` instead of `findIndex` and `slice`

eddiejaoude:
Left some comments. 

Also I noticed when creating a new monitor, the form shows all the form errors for half a second before redirecting the page

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.spec.ts
webkhushboo: okay will remove this file.

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
webkhushboo: done

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
webkhushboo: done

web/src/app/monitors/monitors-list/monitors-list.component.html
webkhushboo: okay done

web/src/app/monitors/monitors-list/monitors-list.component.html
webkhushboo: okay done

web/src/app/projects/view/view.component.html
webkhushboo: done

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
webkhushboo: uid is already added into the form. it was some other issue with add monitor. I have fixed it now

web/src/app/monitors/monitors-list/monitors-list.component.ts
webkhushboo: If I will use filter then it will return me the monitor which I need for update , but I need to update the  array list with the updated monitor value so here logic is 
First I am finding the monitor which I needs to update then using slice I am updating its content.

Please let me know your thoughts on it

webkhushboo:
Fixed this mentioned issue.

web/src/app/core/services/project.service.ts
webkhushboo: sure will do it in that ticket

web/src/app/monitors/monitors-list/monitors-list.component.ts
eddiejaoude: With the filter you can do a not match `!` and it will return all the monitors apart from the one that got deleted.

```typescript
const monitors: MonitorModel[] = this.monitors.filter((monitor: MonitorModel) => monitor.uid !== id);
```

What do you think?

webkhushboo: But I have to update the one in the array

webkhushboo: I have 4 monitors id1, id2, id3 in an array and I have to update monitor with id2. Then how will it possible with filter ?

web/src/app/core/services/project.service.ts
webkhushboo: created monitorService and added saveMonitors to it

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
webkhushboo: done

web/src/app/monitors/monitors-list/monitors-list.component.ts
eddiejaoude: Update? I think this is the delete, so just to return all the ones you want to keep and remove the one you do not?

webkhushboo: ops.. yes in delete this can be done. I have also used the same logic with findIndex and slice for update. So I was thinking about that. I will fix for delete. Thank you :)

web/src/app/core/services/monitor.service.ts
```diff
@@ -0,0 +1,89 @@
...
+   * @param id
+   */
+  findMonitorById(monitors: MonitorModel[], id: string): MonitorModel {
+    return monitors.filter((monitor: MonitorModel) => {
+      return monitor.uid === id;
+    })[0];
+  }
+
+  /**
+   * @param monitors This function is used to save monitor and navigate to monitors list
```
eddiejaoude: Missing parameter in docbloc

web/src/app/core/services/monitor.service.ts
```diff
@@ -0,0 +1,89 @@
...
+  ) {
+  }
+
+  /**
+   *
+   * @param monitors This function will return the monitor by id
+   * @param id
+   */
+  findMonitorById(monitors: MonitorModel[], id: string): MonitorModel {
+    return monitors.filter((monitor: MonitorModel) => {
```
eddiejaoude: Suggestions: 

- this could be simplified to using a `find` and without the `[0]` 
- can be put on a single line without the function `{` and `return`

web/src/app/core/services/monitor.service.ts
```diff
@@ -0,0 +1,89 @@
...
+    private afs: AngularFirestore,
+    private activityService: ActivityService,
+    private snackBar: MatSnackBar,
+    private router: Router
+  ) {
+  }
+
+  /**
+   *
+   * @param monitors This function will return the monitor by id
```
eddiejaoude: I think the method description should be above the parameters not with the first parameter, the text after the parameter is usually the parameter description

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.html
```diff
@@ -0,0 +1,54 @@
+<div *ngIf="monitorForm" class="mat-content" fxLayout="column" fxLayoutGap="20px">
+  <form [formGroup]="monitorForm">
+    <mat-card>
+      <mat-card-header>
+        <mat-card-title> <span *ngIf="!isEdit">Create</span> <span *ngIf="isEdit"> Edit </span> Monitor</mat-card-title>
```
eddiejaoude: Can this be split on multiple lines to be read more easily

web/src/app/monitors/monitors-list/monitors-list.component.html
```diff
@@ -0,0 +1,18 @@
+<mat-card>
+  <div fxLayout="row">
+    <div fxFlex="80%">
+      <h3>Monitors</h3>
+    </div>
+    <div fxFlex="20%"><button mat-raised-button type="submit" color="accent" routerLink="create">Add monitor</button></div>
+  </div>
+  <mat-list>
```
eddiejaoude: Could these 2 sections be split into `mat-card-header` and `mat-card-content` similar to https://github.com/DashboardHub/PipelineDashboard/pull/1332/files#diff-c266124b5a3ac587f336bc3694c09d38

web/src/app/monitors/monitors-list/monitors-list.component.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Lifecycle init method
+   */
+  ngOnInit(): void {
+    this.uid = this.route.snapshot.paramMap.get('uid');
+    this.projectSubscription = this.projectService
+      .findOneById(this.uid)
+      .subscribe((project: ProjectModel) => {
+        this.monitors = project.monitors ? project.monitors : [];
```
eddiejaoude: This could be done on a single line

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.html
```diff
@@ -0,0 +1,54 @@
...
+      <mat-divider></mat-divider>
+
+      <mat-card-content fxLayout="column" fxLayoutGap="20px">
+        <mat-form-field>
+          <input matInput type="text" placeholder="Name" formControlName="name" required>
+          <mat-error *ngIf="monitorForm.get(['name']).hasError('required')">This field is required.</mat-error>
+        </mat-form-field>
+        <mat-form-field>
+          <mat-select formControlName="method" placeholder="Method" required>
+            <mat-option value="get">Get</mat-option>
```
eddiejaoude: The value probably should be capitals to match the model

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
```diff
@@ -0,0 +1,91 @@
...
+   * Lifecycle init method
+   */
+  ngOnInit(): void {
+    this.uid = this.route.snapshot.paramMap.get('uid');
+    this.intializeMonitorForm();
+    this.projectSubscription = this.projectService
+      .findOneById(this.uid)
+      .subscribe((project: ProjectModel) => {
+        this.monitorsList = project.monitors ? project.monitors : [];
+        this.id = this.route.snapshot.paramMap.get('id');
```
eddiejaoude: I think we should always use `uid` because they are both the same format. Can use `projectUid` and `monitorUid` to be clearer

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
```diff
@@ -0,0 +1,91 @@
...
+  /**
+   * This function is used to update the monitor
+   */
+  updateMonitor(): void {
+    const objIndex: number = this.monitorsList.findIndex((monitor: MonitorModel) => monitor.uid === this.id);
+    this.monitorsList = [
+      ...this.monitorsList.slice(0, objIndex),
+      this.monitorForm.value,
+      ...this.monitorsList.slice(objIndex + 1),
+    ];
```
eddiejaoude: I think this block would be clearer if used a negative `filter` and then just pushed the updated monitor on to the array. What do you think?

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
```diff
@@ -0,0 +1,91 @@
...
+    this.uid = this.route.snapshot.paramMap.get('uid');
+    this.intializeMonitorForm();
+    this.projectSubscription = this.projectService
+      .findOneById(this.uid)
+      .subscribe((project: ProjectModel) => {
+        this.monitorsList = project.monitors ? project.monitors : [];
+        this.id = this.route.snapshot.paramMap.get('id');
+        if (this.id) {
+          this.isEdit = true;
+          const filteredMonitor: MonitorModel = this.monitorService.findMonitorById(this.monitorsList, this.id);
```
eddiejaoude: I think this would be better on a `resolver`, sometime I get an empty form then data loads

web/src/app/core/services/monitor.service.ts
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   *
+   * @param uid This function is used to save monitors in the database
+   * @param monitors
+   */
+  public saveMonitors(uid: string, monitors: MonitorModel[]): Observable<void> {
+    if (!monitors.length) {
+      return this.activityService
+        .start()
+        .pipe(
```
eddiejaoude: I sometimes get a error in the console, I think it would be safer to make sure we end this observable as we know we don't need anything else from it, in the `pipe` use `take(1)`

web/src/app/core/services/monitor.service.ts
```diff
@@ -0,0 +1,89 @@
...
+                updatedOn: firebase.firestore.Timestamp.fromDate(new Date()),
+              },
+              { merge: true }))
+        );
+    }
+
+    return this.activityService
+      .start()
+      .pipe(
+        switchMap(() => this.afs
```
eddiejaoude: `take(1)` similar to above

web/src/app/core/services/monitor.service.ts
webkhushboo: What are you referring here ?

web/src/app/core/services/monitor.service.ts
webkhushboo: done

web/src/app/core/services/monitor.service.ts
webkhushboo: right

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.html
webkhushboo: done

web/src/app/monitors/monitors-list/monitors-list.component.html
webkhushboo: sure

web/src/app/monitors/monitors-list/monitors-list.component.ts
webkhushboo: done

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.html
webkhushboo: Which model ?

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
webkhushboo: done

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
webkhushboo: Yes we can do that , but in that case position of updated object will be change in the list . Is that fine then I will made these changes ?

web/src/app/core/services/monitor.service.ts
webkhushboo: sure

web/src/app/core/services/monitor.service.ts
webkhushboo: sure

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
webkhushboo: Resolver is required when we need to check some access on path if we have to navigate or not. 
Please let me know the scenario when you get the empty form and then data loads ?

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
webkhushboo: done

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
eddiejaoude: Good point 👍 . We should sort the list on the UI. Either by date if we have it or alphabetically

webkhushboo: sorting by name ?

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.html
eddiejaoude: The `MonitorModel` has `GET` and `POST` in capitals, the value here should match

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
eddiejaoude: To check access first would be a `guard`, a `resolver` is to load the data before the page loads. In this case, we can load the data before the form is displayed.

The form is sometimes empty with a slow connection on the 2nd visit, then it is populate, a resolve would fix this issue as the form won't display until the data is ready

web/src/app/core/services/monitor.service.ts
eddiejaoude: I have started doing this more when realtime data is not required

web/src/app/core/services/monitor.service.ts
```diff
@@ -0,0 +1,93 @@
...
+    private snackBar: MatSnackBar,
+    private router: Router
+  ) {
+  }
+
+  /**
+   *
+   * @param monitors monitors list
+   * @param id id of monitor which needs to be find
+   * This function will return the monitor by id
```
eddiejaoude: My understanding is the compodoc title is first then parameters.

```compodoc
/**
 * This is a good example
 * 
 * @param {string} target  The target to process see {@link Todo}
 * @returns The processed target number
 */
```

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.html
```diff
@@ -0,0 +1,58 @@
+<div *ngIf="monitorForm" class="mat-content" fxLayout="column" fxLayoutGap="20px">
+  <form [formGroup]="monitorForm">
+    <mat-card>
+      <mat-card-header>
+        <mat-card-title>
+          <span *ngIf="!isEdit">Create</span>
+          <span *ngIf="isEdit"> Edit </span>
```
eddiejaoude: The spaces are not needed around `Edit`

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
eddiejaoude: yep sort by name is fine 👍 - but this can always come as a future enhancement

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.html
webkhushboo: done

web/src/app/core/services/monitor.service.ts
webkhushboo: done

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.html
webkhushboo: done

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
webkhushboo: done

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
```diff
@@ -0,0 +1,87 @@
...
+  ) { }
+
+  /**
+   * Lifecycle init method
+   */
+  ngOnInit(): void {
+    this.projectUid = this.route.snapshot.paramMap.get('projectUid');
+    this.intializeMonitorForm();
+    this.projectSubscription = this.projectService
+      .findOneById(this.projectUid)
```
eddiejaoude: Now the resolver is on the route, it has the information already, so we don't need to fetch it again

webkhushboo: okay done

PR #1343, opened 2019-06-29
feat(project): #1339 added url in project

### LINKED ISSUE #1339: Add url to project
Need to add `url` to project (`project.url`) with validation. Strip off trailing `/`

webkhushboo:
closes #1339 

### Notes

- [ ] Added URL to project
- [ ] Added validation for url 
- [ ] Added pipe for stripe of trailing slash while displaying on my projects screen


web/src/app/shared/components/projects/projects.component.html
```diff
@@ -5,6 +5,7 @@ <h3 mat-subheader>Projects</h3>
     <mat-icon mat-list-icon [matBadge]="project.repositories.length" matBadgeColor="accent" *ngIf="project.type === 'private'">lock</mat-icon>
     <h4 mat-line><a [routerLink]="['/projects', project.uid]">{{ project.title }}</a></h4>
     <p mat-line>{{ project.description }}</p>
+    <p mat-line *ngIf="project.url"><a href="{{project.url | stripeSlash}}">{{ project.url | stripeSlash }}</a></p>
```
eddiejaoude: I think we should strip the slash on the way in not the way out

webkhushboo: If we will add a check while adding the url , then we have to check for the particular field then insert into database. If we have to add at multiple places then we have to do it same everywhere. We can format the URL anywhere in the application using just pipe operator.

eddiejaoude: This is correct if we just want to display it differently. However, the service or model should be the central place to sanitise data, we should not save it incorrectly and then display it correctly 🤓

webkhushboo: okay I will fix it and remove the pipe.

eddiejaoude: Thank you 👍

web/src/app/projects/create-edit/create-edit.component.ts
```diff
@@ -67,4 +69,11 @@ export class CreateEditProjectComponent implements OnInit {
     this.createEditSubscription.unsubscribe();
     this.projectSubscription.unsubscribe();
   }
+
+  /**
+   * Getter for url
+   */
+  get url(): AbstractControl {
+    return this.projectForm.get('url');
+  }
```
eddiejaoude: It doesn't look like this is used anymore? 

`public projectForm: FormGroup;` is public and used in the html `<mat-error *ngIf="projectForm.get(['url']).hasError('pattern')">`

webkhushboo: I have created this method to get the url control and set its value after stripe of the trailing slash

webkhushboo: I think we should always have getters for all properties of form instead of using formgroup everywhere to find that field.
Suppose if we have to check 3 or 4 validation error then we have to write four time 
projectForm.get(['url'])

webkhushboo: moved the stripe call in service from component.

Issue #1124, opened 2018-12-31
Create Repo Webhook when added to project


eddiejaoude:
Dependant on #1285 

**Overview**
When repo is added to a project, create GitHub webhook to send realtime data to PipelineDashboard

**Steps:**
- [x] Check if Webhook already exists, if yes, then do nothing (webhook `id` stored under `webhook` in `repositories` collection for each repo)
- [x] If doesn't exist create GitHub webhook with name `DashboaredHub`

**Notes:**
- [ ] On project dashboard page, if repo has no webhook (`id`), display *warning* message with action link to recreate webhook
- [ ] when repo document being deleted from DB because last usage, must also remove webhook - *delete trigger on cloud function*

GitHub API https://developer.github.com/v3/repos/hooks/

PR #1365, opened 2019-07-15
feat(repositories)#1124 add git webhook to repository

kykyk:
Done:

- alert if there is no webhook on repo was added;
- function for webhook creation using http query was added.

Problems:

- I can't assign a name to webhook
- problem w/ auto removing: I can't get user info from event into trigger (https://stackoverflow.com/questions/46912161/getting-the-user-id-from-a-firestore-trigger-in-cloud-functions-for-firebase). Possible solution: create a cloud function instead of trigger for this.

### DIFF
--- a/functions/package.json
+++ b/functions/package.json
@@ -15,8 +15,10 @@
   "main": "lib/index.js",
   "dependencies": {
     "@google-cloud/logging-winston": "^1.0.0",
-    "firebase-admin": "^7.4.0",
-    "firebase-functions": "^2.3.1",
+    "cors": "^2.8.5",
+    "express": "^4.17.1",
+    "firebase-admin": "^8.2.0",
+    "firebase-functions": "^3.0.2",
     "request": "^2.88.0",
     "request-promise-native": "^1.0.7",
     "winston": "^3.2.1"
--- a/functions/package.json
+++ b/functions/package.json
@@ -16,7 +16,6 @@
   "dependencies": {
     "@google-cloud/logging-winston": "^1.0.0",
     "cors": "^2.8.5",
-    "express": "^4.17.1",
     "firebase-admin": "^8.2.0",
     "firebase-functions": "^3.1.0",
     "request": "^2.88.0",
@@ -25,6 +24,7 @@
     "winston": "^3.2.1"
   },
   "devDependencies": {
+    "@types/cors": "^2.8.5",
     "@types/request-promise-native": "^1.0.16",
     "firebase-functions-test": "^0.1.6",
     "tslint": "~5.16.0",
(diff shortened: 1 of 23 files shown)


eddiejaoude:
> I can't assign a name to webhook

You are right. I saw this in the docs.

`Use web to create a webhook. Default: web. This parameter only accepts the value web.`
https://developer.github.com/v3/repos/hooks/#create-a-hook

My suggestion, when creating the webhook for the first time, we can save the `uid` in the repository document and then can use that.

functions/package.json
```diff
@@ -15,6 +15,8 @@
   "main": "lib/index.js",
   "dependencies": {
     "@google-cloud/logging-winston": "^1.0.0",
+    "cors": "^2.8.5",
+    "express": "^4.17.1",
```
eddiejaoude: Do we really need express? Seems a bit overkill

functions/src/repository/create-git-webhook-repository.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export async function getWebhook(repositoryUid: string, token: string): Promise<GitHubRepositoryWebhookModel> {
+
+  const exist: GitHubRepositoryWebhookResponse = await findWebhook(repositoryUid, token);
+
+  if (exist) {
+    return GitHubRepositoryWebhookMapper.import(exist);
+  } else {
+    return GitHubRepositoryWebhookMapper.import(await createWebhook(repositoryUid, token));
+  }
```
eddiejaoude: Not important, I just think its cleaner to have it written like this...

```typescript
if (exist) {
  return GitHubRepositoryWebhookMapper.import(exist);
}
     
return GitHubRepositoryWebhookMapper.import(await createWebhook(repositoryUid, token));
```

functions/src/repository/response-git-webhook-repository.ts
```diff
@@ -0,0 +1,21 @@
...
+});
+
+export interface ResponseGitWebhookRepositoryInput {
+  token: string;
+  repositoryUid: string;
+}
+
+export const onResponseGitWebhookRepository: HttpsFunction = https.onRequest((req: https.Request, res: Response) => {
+  return cors(req, res, () => {
+    console.log(`${req.protocol}://${req.hostname} ; onResponseGitWebhookRepository: success!`);
```
eddiejaoude: Is this `console` needed? If so can we use the `Logger`

eddiejaoude:
When resolving the conflicts, I broke the build. I will fix it 🤦‍♂

eddiejaoude:
Thank you for fixing 👍

PR #1374, opened 2019-07-24
feat(monitors): #1367 delete pings subcollection on deleting project

### LINKED ISSUE #1367: When delete project delete all pings
Delete all sub collections. Currently only `pings`, don't get monitor uids, just delete all `pings` under the `project`

webkhushboo:
closes #1367 

### Notes

- [ ] Added deleteProjectPings function for deleting the pings sub collection when deleting the project

### DIFF
--- a/functions/src/index.ts
+++ b/functions/src/index.ts
@@ -2,7 +2,9 @@ import * as functions from 'firebase-functions';
 import { CallableContext } from 'firebase-functions/lib/providers/https';
 import { getRepositoryInfo, RepositoryInfoInput } from './repository/info';
 
-import { deleteMonitorPings, ping, MonitorInfoInput } from './monitor/monitor';
+import { DocumentSnapshot } from 'firebase-functions/lib/providers/firestore';
+import { ProjectModel } from './models/index.model';
+import { deleteMonitorPings, deleteProjectPings, ping, MonitorInfoInput } from './monitor/monitor';
 import { getUserEvents, EventsInput } from './user/events';
 import { getUserRepos, ReposInput } from './user/repos';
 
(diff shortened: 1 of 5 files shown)


webkhushboo:
@eddiejaoude 

While checking the function logs , I am seing the below error 
`Function execution took 79 ms, finished with status: 'error'`
Although function works fine. Can you please provide your inputs on it?

eddiejaoude:
Looks good. I will test it out tomorrow. 

Can you update with the main branch and resolve conflicts

webkhushboo:
@eddiejaoude 
I have fixed the above issue and resolved the conflicts too
Kindly review

functions/src/project/delete-project.ts
```diff
@@ -40,3 +41,27 @@ export const onDeleteProjectRepositories: CloudFunction<DocumentSnapshot> = fire
...
+  .document('projects/{projectUid}')
+  .onDelete(async (projectSnapshot: DocumentSnapshot, context: EventContext) => {
+
+    try {
+      const project: DocumentData = projectSnapshot.data();
+
+      if (Array.isArray(project.monitors) && project.monitors.length > 0) {
+        const promises: Promise<WriteResult>[] = [];
+
+        for (const monitor of project.monitors) {
```
eddiejaoude: I think this can be all removed if instead of using `deleteMonitorPings`, `deleteProjectPings` is used because it has the loop already?

functions/src/project/delete-project.ts
```diff
@@ -40,3 +41,27 @@ export const onDeleteProjectRepositories: CloudFunction<DocumentSnapshot> = fire
...
+          promises.push(deleteMonitorPings(project.uid, monitor.uid))
+        }
+        await Promise.all(promises);
```
eddiejaoude: I think we can return the `Promise.all`?

eddiejaoude: Some places we do and some we don't. Let me check

eddiejaoude: > Note: If a Node.js background function returns a Promise, Cloud Functions ensures that the Promise is settled before terminating.

I think it is better if we always return a promise. I will try to update other functions later on in another PR, but if this one can be updated in this PR please 🤓

functions/src/project/delete-project.ts
webkhushboo: Good catch .

functions/src/project/delete-project.ts
webkhushboo: This will be removed as we will call deleteProjectPings here

eddiejaoude:
This is being closed as it is now part of #1375

PR #1377, opened 2019-07-28
feat(pings): #1353 added function to run all monitors every 60 mins

### LINKED ISSUE #1353: Run all monitors every 60mins
Cloud function on a cron to run every 60mins to get all monitors on projects and run their monitors then to call  #1350 for each monitor

*Note: Use project url to prepend monitor path for full url*

- cloud function every 60mins
- find all projects who have `monitors` (new 2019 feature Firestore to do this?)

webkhushboo:
closes #1353 

### Notes

- [ ] API: Created cloud function to run all monitors of project every 60 mins

### DIFF
--- a/functions/src/scheduler/schedule.ts
+++ b/functions/src/scheduler/schedule.ts
@@ -41,9 +41,8 @@ export const runAllMonitors60Mins: any = functions.pubsub.schedule('every 60 min
           for (const monitor of project.monitors) {
             promises.push(ping(project.uid, monitor.uid))
           }
-          await Promise.all(promises);
         }
-        await Promise.all(promises);
+        return Promise.all(promises);
 
       } catch (err) {
         Logger.error(err);
--- a/functions/src/scheduler/schedule.ts
+++ b/functions/src/scheduler/schedule.ts
@@ -34,16 +34,16 @@ export const runAllMonitors60Mins: any = functions.pubsub.schedule('every 60 min
     
     snapshots.docs.forEach(async (doc: firestore.QueryDocumentSnapshot) => {
       const project: ProjectModel = <ProjectModel>doc.data();
+      const promises: Promise<WriteResult>[] = [];
       try {
         if (project.url && Array.isArray(project.monitors) && project.monitors.length > 0) {
-          const promises: Promise<WriteResult>[] = [];
 
           for (const monitor of project.monitors) {
             promises.push(ping(project.uid, monitor.uid))
           }
           await Promise.all(promises);
         }
-        await Promise.all([]);
+        await Promise.all(promises);
 
       } catch (err) {
         Logger.error(err);
--- a/functions/src/scheduler/schedule.ts
+++ b/functions/src/scheduler/schedule.ts
@@ -5,7 +5,7 @@ import { FirebaseAdmin } from '../client/firebase-admin';
 
 // Dashboard Hub models
 import { Logger } from '../client/logger';
-import { ProjectModel } from '../models/index.model';
+import { MonitorModel, ProjectModel } from '../models/index.model';
 import { ping } from '../monitor/monitor';
 import { deleteProjectPings } from '../project/delete-project';
 
@@ -32,21 +32,18 @@ export const runAllMonitors60Mins: any = functions.pubsub.schedule('every 60 min
       .collection('projects')
       .get();
     
+    const promises: Promise<WriteResult>[] = [];  
+    let project: ProjectModel;
     snapshots.docs.forEach(async (doc: firestore.QueryDocumentSnapshot) => {
-      const project: ProjectModel = <ProjectModel>doc.data();
-      const promises: Promise<WriteResult>[] = [];
+      project = <ProjectModel>doc.data();
       try {
         if (project.url && Array.isArray(project.monitors) && project.monitors.length > 0) {
-
-          for (const monitor of project.monitors) {
-            promises.push(ping(project.uid, monitor.uid))
-          }
+          project.monitors.forEach((monitor: MonitorModel) => promises.push(ping(project.uid, monitor.uid)));
         }
-        return Promise.all(promises);
-
       } catch (err) {
         Logger.error(err);
         throw new Error(err);
       }
     });
+    return Promise.all(promises);
   });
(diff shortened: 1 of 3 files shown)


functions/src/scheduler/schedule.ts
```diff
@@ -23,3 +25,29 @@ export const deletePingsAfter30days: any = functions.pubsub.schedule('00 00 * *
     });
     return Promise.all(promises);
   });
+
+export const runAllMonitors60Mins: any = functions.pubsub.schedule('every 60 mins')
+  .onRun(async () => {
+    const snapshots: QuerySnapshot = await FirebaseAdmin.firestore()
+      .collection('projects')
```
eddiejaoude: I think we should filter this out for projects that have `url`s

functions/src/scheduler/schedule.ts
```diff
@@ -23,3 +25,29 @@ export const deletePingsAfter30days: any = functions.pubsub.schedule('00 00 * *
...
+      try {
+        if (project.url && Array.isArray(project.monitors) && project.monitors.length > 0) {
+          const promises: Promise<WriteResult>[] = [];
+
+          for (const monitor of project.monitors) {
+            promises.push(ping(project.uid, monitor.uid))
+          }
+          await Promise.all(promises);
+        }
+        await Promise.all([]);
```
eddiejaoude: I think instead of creating a new `Promise.all([])`, we should return the `Promise.all(promises)` which could be an empty array

eddiejaoude: My thoughts:

- define the `promises` outside the `try` 
- return `Promise.all(promises)` after the `try/catch`

webkhushboo: Sure!!

functions/src/scheduler/schedule.ts
webkhushboo: Firestore does not provide `!=` in where clause.
Also there is already a check added for proejct.url before calling `ping` function

functions/src/scheduler/schedule.ts
```diff
@@ -23,3 +25,29 @@ export const deletePingsAfter30days: any = functions.pubsub.schedule('00 00 * *
...
+    snapshots.docs.forEach(async (doc: firestore.QueryDocumentSnapshot) => {
+      const project: ProjectModel = <ProjectModel>doc.data();
+      const promises: Promise<WriteResult>[] = [];
+      try {
+        if (project.url && Array.isArray(project.monitors) && project.monitors.length > 0) {
+
+          for (const monitor of project.monitors) {
+            promises.push(ping(project.uid, monitor.uid))
+          }
+          await Promise.all(promises);
```
eddiejaoude: I don't think this promise all is required?

functions/src/scheduler/schedule.ts
```diff
@@ -23,3 +25,29 @@ export const deletePingsAfter30days: any = functions.pubsub.schedule('00 00 * *
...
+        }
+        await Promise.all(promises);
```
eddiejaoude: I think this could be a `return`

functions/src/scheduler/schedule.ts
```diff
@@ -23,3 +25,28 @@ export const deletePingsAfter30days: any = functions.pubsub.schedule('00 00 * *
...
+      const project: ProjectModel = <ProjectModel>doc.data();
+      const promises: Promise<WriteResult>[] = [];
+      try {
+        if (project.url && Array.isArray(project.monitors) && project.monitors.length > 0) {
+
+          for (const monitor of project.monitors) {
+            promises.push(ping(project.uid, monitor.uid))
+          }
+        }
+        return Promise.all(promises);
```
eddiejaoude: I prefer the return outside the `try`, I can move this before I merge

functions/src/scheduler/schedule.ts
```diff
@@ -23,3 +25,28 @@ export const deletePingsAfter30days: any = functions.pubsub.schedule('00 00 * *
...
+
+export const runAllMonitors60Mins: any = functions.pubsub.schedule('every 60 mins')
+  .onRun(async () => {
+    const snapshots: QuerySnapshot = await FirebaseAdmin.firestore()
+      .collection('projects')
+      .get();
+    
+    snapshots.docs.forEach(async (doc: firestore.QueryDocumentSnapshot) => {
+      const project: ProjectModel = <ProjectModel>doc.data();
+      const promises: Promise<WriteResult>[] = [];
```
eddiejaoude: Should this not be outside the `forEach` loop?

functions/src/scheduler/schedule.ts
```diff
@@ -23,3 +25,28 @@ export const deletePingsAfter30days: any = functions.pubsub.schedule('00 00 * *
...
   });
+
+export const runAllMonitors60Mins: any = functions.pubsub.schedule('every 60 mins')
+  .onRun(async () => {
+    const snapshots: QuerySnapshot = await FirebaseAdmin.firestore()
+      .collection('projects')
+      .get();
+    
+    snapshots.docs.forEach(async (doc: firestore.QueryDocumentSnapshot) => {
+      const project: ProjectModel = <ProjectModel>doc.data();
```
eddiejaoude: I dont think we want to redefine the `const` on each loop

functions/src/scheduler/schedule.ts
```diff
@@ -23,3 +25,28 @@ export const deletePingsAfter30days: any = functions.pubsub.schedule('00 00 * *
...
+      const promises: Promise<WriteResult>[] = [];
+      try {
+        if (project.url && Array.isArray(project.monitors) && project.monitors.length > 0) {
+
+          for (const monitor of project.monitors) {
```
eddiejaoude: I prefer `forEach` if possible to be consistent

functions/src/scheduler/schedule.ts
webkhushboo: We are creating list of promises for each project so we have to create promise array inside the loop. If I will defined outside the loop then it will push all

functions/src/scheduler/schedule.ts
webkhushboo: done

functions/src/scheduler/schedule.ts
eddiejaoude: Promise all runs in parallel, so I think creating a big list is ok?

functions/src/scheduler/schedule.ts
webkhushboo: done

functions/src/scheduler/schedule.ts
webkhushboo: Done

PR #1381, opened 2019-08-01
feat(monitors): #1379 added type in ping model

### LINKED ISSUE #1379: Ping type "scheduler" or "manual"
Can we add a `type` to the `ping` result, so that it is clear if it was a scheduler or a manual ping result

webkhushboo:
closes #1379 

### Notes

- [ ] API: Added ping type in ping function
- [ ] UI: Send "manual" from UI while ping monitor
- [ ] API: Set "scheduler" type if ping by scheduler

### DIFF
--- a/functions/src/models/ping.model.ts
+++ b/functions/src/models/ping.model.ts
@@ -15,4 +15,5 @@ export interface PingModel {
   statusCode: number;
   textMatched: boolean;
   createdOn: firestore.Timestamp;
+  type: string;
 }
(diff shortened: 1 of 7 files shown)


functions/src/models/ping.model.ts
```diff
@@ -15,4 +15,5 @@ export interface PingModel {
   statusCode: number;
   textMatched: boolean;
   createdOn: firestore.Timestamp;
+  type: string;
```
eddiejaoude: I am going to change this to be a bit more strict

```typescript
type: 'scheduler' | `manual`;
```

Issue #1123, opened 2018-12-31
Realtime events - parse webhooks from GitHub


eddiejaoude:
Dependant on #1124 

To have data updated in realtime from GitHub, create a Webhooks and cloud function that catches and maps data to Firestore.

Pick 1 or 2 webhook events, something we would see on the dashboard like `issues` / `pull requests` / `events` - don't map all the data, just the ones we currently use

PR #1397, opened 2019-08-08
feat(repositories)#1123 parse webhooks from GitHub

kykyk:
Done:
- models for response of webhook
- logic of detect the event type
- parse events

Problem:
- update contributors (github cache https://developer.github.com/v3/repos/statistics/)


eddiejaoude:
Lint error on CI 

```
ERROR in src/app/projects/repository/repository.component.ts(65,43): error TS2345: Argument of type 'string' is not assignable to parameter of type 'RepositoryModel'.
```

eddiejaoude:
I have created an issue for the contributors data, as this is low priority and important for now  https://github.com/DashboardHub/PipelineDashboard/issues/1460

functions/src/environments/environment.ts
```diff
@@ -1,10 +1,58 @@
 
 export const enviroment: Config = {
   githubWebhook: {
-    url: "https://us-central1-pipelinedashboard-dev-cfa68.cloudfunctions.net/responseGitWebhookRepository",
+    url: "https://us-central1-project-pipeline-15611.cloudfunctions.net/responseGitWebhookRepository",
```
eddiejaoude: I need to move this to CI to change depending on the deployment or local build - this is a note for myself 🤓

functions/src/repository/response-git-webhook-repository.ts
```diff
@@ -17,6 +34,199 @@ export interface ResponseGitWebhookRepositoryInput {
...
     Logger.info(`${req.protocol}://${req.hostname} ; onResponseGitWebhookRepository: success!`);
-    res.status(200).send();
+
+    const inputData: any = req.body;
+    let result: Promise<any>;
+
+    Logger.info(req.body);
+    Logger.info(Object.keys(inputData));
+
+    if (IssueCommentEventModel.isCurrentModel(inputData)) {
```
eddiejaoude: This would be better as a `switch`

functions/src/repository/response-git-webhook-repository.ts
```diff
@@ -17,6 +34,199 @@ export interface ResponseGitWebhookRepositoryInput {
 export const onResponseGitWebhookRepository: HttpsFunction = https.onRequest((req: https.Request, res: Response) => {
   return cors(req, res, () => {
     Logger.info(`${req.protocol}://${req.hostname} ; onResponseGitWebhookRepository: success!`);
-    res.status(200).send();
+
+    const inputData: any = req.body;
+    let result: Promise<any>;
+
+    Logger.info(req.body);
+    Logger.info(Object.keys(inputData));
```
eddiejaoude: `req.body` and `inputData` are the same. If we need both loggers, we should be consistent in the variable we use to reference the same data. I guess `inputData` makes more sense at this stage in the code

PR #1405, opened 2019-08-12
feat(monitors): #1403 update monitor and send automatic ping

### LINKED ISSUE #1403: When monitor updated, send automatic manual ping
When a monitor is updated, send a manual ping automatically

May be with another type like `automatic`

webkhushboo:
closes #1403 

### Notes

- [ ] Updated monitor with existing ping info
- [ ] Sent automatic ping while updating monitor


web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
```diff
@@ -96,7 +96,10 @@ export class MonitorCreateEditComponent implements OnInit, OnDestroy {
   saveMonitor(uid: string, monitors: MonitorModel[]): void {
     this.saveMonitorSubscription = this.monitorService.saveMonitors(uid, monitors)
       .subscribe(
-        () => this.router.navigate([`/projects/${uid}/monitors`]),
+        () => {
+          this.monitorService.pingMonitor(this.projectUid, this.monitorUid, 'automatic');
```
eddiejaoude: this should probably go in a `pipe` rather than the `subscribe`

webkhushboo: fixed it , added a check for save and update monitor scenario too.

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
```diff
@@ -93,20 +94,36 @@ export class MonitorCreateEditComponent implements OnInit, OnDestroy {
...
    *
    */
-  saveMonitor(uid: string, monitors: MonitorModel[]): void {
-    this.saveMonitorSubscription = this.monitorService.saveMonitors(uid, monitors)
-      .subscribe(
-        () => this.router.navigate([`/projects/${uid}/monitors`]),
-        (error: any): any => this.snackBar.open(error.message, undefined, { duration: 5000 })
-      );
+  saveMonitor(uid: string, monitors: MonitorModel[], monitorUid: string): void {
+    if (monitorUid) {
```
eddiejaoude: I don't think we need the `monitorUid` because it is available in `this.monitorUid`?

eddiejaoude: I think this would be simpler if we moved the `automatic` ping to the service `saveMonitors`, then the if/else is no longer required here?

web/src/app/shared/models/ping.model.ts
```diff
@@ -1,3 +1,5 @@
+export type PingType = 'scheduler' | 'manual'| 'automatic';
```
eddiejaoude: 👍

web/src/app/monitors/monitor-create-edit/monitor-create-edit.component.ts
webkhushboo: > I don't think we need the `monitorUid` because it is available in `this.monitorUid`?

Fixed

webkhushboo: > I think this would be simpler if we moved the `automatic` ping to the service `saveMonitors`, then the if/else is no longer required here?

We need to call the `pingMonitor()` with type `automatic` when monitor is updated so I think it should be added here in `saveMonitors()subscribe()`.

webkhushboo: Fixed and called the `pingMonitor` for `automatic update` inside the service

web/src/app/monitors/monitors-list/monitors-list.component.ts
```diff
@@ -84,7 +84,7 @@ export class MonitorsListComponent implements OnInit, OnDestroy {
   public pingMonitor(monitorUid: string): void {
     this.manualPing = true;
     this.monitorService
-      .pingMonitor(this.projectUid, monitorUid);
+      .pingMonitor(this.projectUid, monitorUid, 'manual');
```
eddiejaoude: I just noticed this needs a subscribe because its an observable

web/src/app/core/services/monitor.service.ts
```diff
@@ -43,9 +43,9 @@ export class MonitorService {
           if (monitors.find((monitorModel: MonitorModel) => monitorModel.uid === monitor.uid)) {
             return project.monitors.map((monitorModel: MonitorModel) => {
               if (monitorModel.uid === monitor.uid) {
+                this.pingMonitor(projectUid, monitor.uid, 'automatic');
```
eddiejaoude: I just noticed this needs to be returned in a the `pipe` because its an observable

web/src/app/monitors/monitors-list/monitors-list.component.ts
eddiejaoude: I have had a play with this and will push up my changes for you to review

web/src/app/core/services/monitor.service.ts
eddiejaoude: this one was a bit interesting, I have had a play with this and will push up my changes for you to review

PR #1441, opened 2019-08-29
feat(error): #1422 Error module and 404 page

### LINKED ISSUE #1422: Error page (404)
Create error module and catch all UI routing that is not matched and create a `404` page using the mascot with ears down...

eddiejaoude:
closes #1422 

### Notes

- [ ] Error module
- [ ] 404 error page

Screenshot 2019-08-29 08 17 17

### DIFF
--- a/web/src/app/error/error.module.ts
+++ b/web/src/app/error/error.module.ts
@@ -1,0 +1,18 @@
+// Core components
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+
+// Application component
+import { SharedModule } from '@app/shared/shared.module';
+import { ErrorRoutingModule } from './error-routing.module';
+import { ErrorComponent } from './error.component';
+
+@NgModule({
+  declarations: [ErrorComponent],
+  imports: [
+    CommonModule,
+    ErrorRoutingModule,
+    SharedModule,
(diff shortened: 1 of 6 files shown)


web/src/app/error/error.module.ts
```diff
@@ -0,0 +1,18 @@
+// Core components
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+
+// Application component
+import { SharedModule } from '@app/shared/shared.module';
```
webkhushboo: Here we can use `@shared` instead of `@app/shared`

eddiejaoude: Good spot 👍

eddiejaoude: I just copied that part 😂 😜 - I didn't spot it 🤦‍♂ . I will update

eddiejaoude:
Great spot on the scroll. I think I fixed it. Please review again 🤓

web/src/app/error/error.module.ts
webkhushboo: Yes it was my mistake :( . I noticed now so I commented

webkhushboo:
Looks good now :)

PR #1453, opened 2019-09-01
feat(functions): #1446 functions for application stats added

### LINKED ISSUE #1446: Application statistics
Overview for application statistics. New collection `platform` with a single document, unless there is another way?

- project total - update each time a project is created ~~/ deleted~~
- ~~monitor total - update each time a project.monitors is added ~~/removed~~ (same function trigger as project above)~~
- ping total - update each time a ping is created
- user total - update each time a user is created
- GitHub event totals - update each time an event is created (dependant on #1445, so could come in another PR)

For example
```typescript
{
   projects: number;
   pings: number;
   users: number;
   events: number;
}
```

*Note: only increase numbers, dont worry about decreasing them*

webkhushboo:
closes #1446 

### Notes

- [ ] API: Cloud function added for updating project count in database
- [ ] API: Cloud function added for updating users count in database
- [ ] API: Cloud function added for updating pings count in database

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -43,6 +43,13 @@ service cloud.firestore {
       allow update: if false;
       allow delete: if false;
     }
+
+    match /platform/stats {
+      allow create: if false;
+      allow get: if false;
+      allow update: if false;
+      allow delete: if false;
+    }
   }
 
 }
--- a/firestore.rules
+++ b/firestore.rules
@@ -46,7 +46,8 @@ service cloud.firestore {
 
     match /platform/stats {
       allow create: if false;
-      allow get: if false;
+      allow get: if true;
+      allow list: if false;
       allow update: if false;
       allow delete: if false;
     }
--- a/functions/src/application/stats.ts
+++ b/functions/src/application/stats.ts
@@ -1,0 +1,65 @@
+// Third party modules
+import * as firebase from 'firebase-admin';
+import { firestore, CloudFunction, EventContext } from 'firebase-functions';
+import { DocumentSnapshot, FirebaseAdmin } from '../client/firebase-admin';
+import { Logger } from '../client/logger';
+
+export const onCreateProject: CloudFunction<DocumentSnapshot> = firestore
+  .document('projects/{projectUid}')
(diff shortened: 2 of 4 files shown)


firestore.rules
```diff
@@ -43,6 +43,13 @@ service cloud.firestore {
       allow update: if false;
       allow delete: if false;
     }
+
+    match /platform/stats {
+      allow create: if false;
+      allow get: if false;
```
eddiejaoude: I think for `get` it should be for everyone to read it. Also missing list, which I think we should not allow for the moment.

```suggestion
      allow get: if true;
      allow list: if false;
```

functions/src/application/stats.ts
```diff
@@ -0,0 +1,63 @@
+// Third party modules
+import * as firebase from 'firebase-admin';
```
eddiejaoude: I think we can use `functions/src/client/firebase-admin.ts` to make it easier. There might be some missing `export`s that need to be added

eddiejaoude:
Looks good 👍 , left a couple of small comments

functions/src/application/stats.ts
```diff
@@ -0,0 +1,63 @@
...
+          { merge: true });
+    } catch (err) {
+      Logger.error(err);
+      throw new Error(err);
+    }
+  });
+
+
+export const onCreatePings: CloudFunction<DocumentSnapshot> = firestore
+  .document('projects/{projectUid}/pings/{pingUid}')
```
eddiejaoude: I did not know this was possible 👍

functions/src/application/stats.ts
```diff
@@ -0,0 +1,63 @@
...
+          },
+          { merge: true });
+    } catch (err) {
+      Logger.error(err);
+      throw new Error(err);
+    }
+  });
+
+
+export const onCreatePings: CloudFunction<DocumentSnapshot> = firestore
```
eddiejaoude: Extra blank line not needed

functions/src/application/stats.ts
eddiejaoude: I was thinking this was part of the document, but it is a sub collection 🤦‍♂

PR #1470, opened 2019-09-07
feat(projects): #1459 repo best practices rating calculation

### LINKED ISSUE #1459: Repo best practices rating
For each repo we should calculate in the UI how "best practice" it is. We will add & improve the checks over time.

Initial checks *(do the ones we already have data for)*
- Was the last issue created in the last:
   - 1 day = 100%
   - ...
   - 10 day = ((30 - 10) / 100) * 30
   - ...
   - 30 days = 0%
- Was there a release in the last:
   - same as above
- Calculate the overall percentage (the calculation should probably be in the `repositoryService` and the method returns a `percentage`)

Rough example for each data item, total up all items (I think only 2 for now) and divide by total items
```typescript
// let lastIssueDays: number = 10 // example data
const boundary: number = 30 // days

return ((boundary - lastIssueDays) / 30) * 100 // percentage 
```

Total percentage example:
Issues = 40%, Milestone = 60%, then add them together and divide by 2 (I think) 😂 

---

Future ideas / additional check ideas: **(will need more data)**
- Was the last commit in the last:
   - 7 days = 10 points
   - 30 days = 5 points
- Does it have `README.md`, `CONTRIBUTORS.MD`
- Contributors

webkhushboo:
closes #1459 

### Notes

- [ ] Added calculation for repo rating based upon on last release and last issues created on


web/src/app/core/services/repository.service.ts
```diff
@@ -51,4 +51,27 @@ export class RepositoryService {
     const callable: any = this.fns.httpsCallable('deleteGitWebhookRepository');
     return callable({ repositoryUid, token: this.authService.profile.oauth.githubToken });
   }
+
+  public getRepoRating(repo: RepositoryModel): number {
```
eddiejaoude: Because already in `repository` service, `getRating` should be descriptive enough, as it would read `respositoryService.getRating(...)`

web/src/app/core/services/repository.service.ts
```diff
@@ -51,4 +51,27 @@ export class RepositoryService {
...
+    let rating: number = 0;
+    const issuePoints: number = repo.issues.length > 0 ? this.getPoints(repo.issues[0].createdOn) : 0;
+    const releasesPoints: number = repo.releases.length > 0 ? this.getPoints(repo.releases[0].createdOn) : 0;
+    rating = (issuePoints + releasesPoints) * 5;
```
eddiejaoude: Need to calculate percentage, so when more rating calculations are added, it is always out of 100%.

I think here should be divide by the amount (`2` at the moment), and below `getPoints` returns a percentage out of `100`

web/src/app/core/services/repository.service.ts
```diff
@@ -51,4 +51,27 @@ export class RepositoryService {
...
+    return rating;
+  }
+
+  public getPoints(dateStr: Date): number {
```
eddiejaoude: `dateStr: Date`, the variable name implies `string` but the type states `Date`. Better to just use the type and not have it as part of the variable name

web/src/app/core/services/repository.service.ts
```diff
@@ -51,4 +51,27 @@ export class RepositoryService {
...
+    const date: Date = new Date();
+    const referenceDate: Date = new Date(dateStr);
+    let lapse: number = Math.floor((date.getTime() - referenceDate.getTime()) / 1000);
+    const day: number = 24 * 60 * 60;
+    const duration: number = lapse / day;
+    if (duration < 7) {
```
eddiejaoude: A `switch` might be better here

web/src/app/core/services/repository.service.ts
```diff
@@ -51,4 +51,27 @@ export class RepositoryService {
...
+      return 10;
+    } else if (duration < 30) {
+      return 5;
+    } else {
+      return 1;
+    }
```
eddiejaoude: `return` a percentage, `(value / 10) * 100`

web/src/app/projects/repository/repository.component.ts
```diff
@@ -66,6 +68,10 @@ export class RepositoryComponent implements OnInit, OnDestroy {
       .subscribe(() => setTimeout(() => this.manualReload = false, 60000)); // disable the ping button for 60 seconds;
   }
 
+  getRepoRating(): void {
```
eddiejaoude: Already in `repository` component, so name should be fine as `getRating`.

But as there is no `return`, `get` seems odd, better to have something like `calculateRating`

web/src/app/shared/models/index.model.ts
```diff
@@ -3,6 +3,7 @@ export * from './model.model';
 export * from './activity.model';
 export * from './contributor.model';
 export * from './event.model';
+export * from './issue.model';
```
eddiejaoude: I don't see an `issue.model` file?

web/src/app/core/services/repository.service.ts
webkhushboo: Sure !!

web/src/app/core/services/repository.service.ts
webkhushboo: sure !!

web/src/app/shared/models/index.model.ts
webkhushboo: Actually its not part of this PR . `issue.model` was missing from `index.model.ts` file. So I have added here

eddiejaoude: Oh ok, nice spot, thank you 👍

webkhushboo: :)

web/src/app/core/services/repository.service.ts
webkhushboo: done

web/src/app/core/services/repository.service.ts
webkhushboo: calculation logic changed so we don't need the switch here

web/src/app/core/services/repository.service.ts
webkhushboo: done

web/src/app/projects/repository/repository.component.ts
webkhushboo: done

web/src/app/core/services/repository.service.ts
```diff
@@ -51,4 +51,22 @@ export class RepositoryService {
     const callable: any = this.fns.httpsCallable('deleteGitWebhookRepository');
     return callable({ repositoryUid, token: this.authService.profile.oauth.githubToken });
   }
+
+  public getRating(repo: RepositoryModel): number {
```
eddiejaoude: missing docs

web/src/app/projects/repository/repository.component.html
```diff
@@ -17,6 +17,9 @@
   <mat-expansion-panel hideToggle #expansion>
     <mat-expansion-panel-header [collapsedHeight]="headerHeight+'px'" [expandedHeight]="headerHeight+'px'" fxLayout="row" fxLayout.lt-lg="column" fxLayoutAlign="space-between start">
       <div fxLayout="column" class="repository__container">
+        <div fxLayout="row">
+            <mat-progress-bar matTooltipPosition="above" matTooltip="Rating: {{rating | number:'1.0-2'}}%" mode="determinate" value={{rating}}></mat-progress-bar>
```
eddiejaoude: Missing spacing `{{rating | number:'1.0-2'}}` to be `{{ rating | number:'1.0-2' }}`

eddiejaoude: Value needs spacing but better with quotes and brackets `[value]="rating"`

eddiejaoude:
If no data, get a negative percentage

Screenshot 2019-09-08 19 43 07

web/src/app/projects/repository/repository.component.scss
```diff
@@ -330,3 +330,8 @@
   }
 
 }
+
+.rating{
```
eddiejaoude: Needs a space `.rating {`

web/src/app/projects/repository/repository.component.scss
```diff
@@ -330,3 +330,8 @@
...
+  padding-left: 10px;
+  font-size: 16px;
+}
```
eddiejaoude: This does not seem to be used?

web/src/app/core/services/repository.service.ts
```diff
@@ -51,4 +51,22 @@ export class RepositoryService {
...
+    rating = (issuePoints + releasesPoints) / 2;
+    return rating;
+  }
+
+  public getPoints(date: Date): number {
+    const currentDate: Date = new Date();
+    const referenceDate: Date = new Date(date);
+    let lapse: number = Math.floor((currentDate.getTime() - referenceDate.getTime()) / 1000);
+    const hoursInDay: number = 24 * 60 * 60;
+    const duration: number = Math.ceil(lapse / hoursInDay);
```
eddiejaoude: If greater than `boundary`, then should be `0` otherwise goes negative

eddiejaoude: I will resolve 🤓

web/src/app/projects/repository/repository.component.html
eddiejaoude: I will resolve 🤓

web/src/app/projects/repository/repository.component.scss
eddiejaoude: I will remove 🤓

Issue #1411, opened 2019-08-16
Key of repository collection should be changed to generated UID


kykyk:
If user wants to rename repo, we should change key of repos collection in database as it consists of username and repo name, but it is impossible.

- [x] generate key UID automatically
- [x] change logic of CRUD operations and queries in 'web' and 'function'

eddiejaoude:
One solution to this, is they remove it from the project and re-add it again, but you are right it is not ideal because everyone will need to do this and then Webhooks would have to be re-created again.

Does the repo rename come in via a webhook event?

kykyk:
> One solution to this, is they remove it from the project and re-add it again, but you are right it is not ideal because everyone will need to do this and then Webhooks would have to be re-created again.
> 
> Does the repo rename come in via a webhook event?

Webhook has a RepositoryEvent, this event includes a 'rename' action
https://developer.github.com/v3/activity/events/types/#repositoryevent

eddiejaoude:
> this event includes a 'rename' action

Ok great 👍

PR #1485, opened 2019-09-09
feat(repositories)#1411 key of repository collection

kykyk:
Done:
- generate key UID automatically
- change logic of CRUD operations and queries in 'web' and 'function'
- add function to avoid duplicate repository in collection

What should be done:
- deploy functions
- clear database and browser cache


functions/src/models/repository.model.ts
```diff
@@ -0,0 +1,45 @@
...
+export class RepositoryModel implements IRepository {
+  id: number;
+  fullName: string;
+  uid?: string;
+
+  public static getRepositoryReference(uid: string): DocumentReference {
+    return FirebaseAdmin.firestore().collection('repositories').doc(uid);
+  }
+
+  public static async getRepositoryById(id: number): Promise<DocumentData | null> {
```
eddiejaoude: These looks more like `service` methods because the *model* knows the `id` already. I will look into refactoring this

functions/src/project/delete-project.ts
```diff
@@ -3,67 +3,67 @@ import { firestore, CloudFunction, EventContext } from 'firebase-functions';
...
+    if (Array.isArray(project.repositories) && project.repositories.length > 0) {
+      for (const repositoryUid of project.repositories) {
+        if (!repositoryUid) {
+          continue;
         }
+        const repoRef: DocumentReference = RepositoryModel.getRepositoryReference(repositoryUid);
 
-        await Promise.all(promiseList);
-      }
+        await FirebaseAdmin.firestore().runTransaction((t: Transaction) => {
```
eddiejaoude: Missing `return`, the function return type is expecting `promise` return. I will refactor

functions/src/repository/create-repository.ts
```diff
@@ -16,23 +16,82 @@ async function addProjects(repositoryUid: string, repo: DocumentData): Promise<b
...
+                t.update(element.ref, { repositories: uids });
+              }
+            }
+          });
+
+        });
+    });
+
+    await repoRef.delete();
+    return false;
```
eddiejaoude: Function return type is `promise` but `boolean` here. I will fix

eddiejaoude:
Does not seem to work?

Screenshot 2019-09-10 07 52 42

kykyk:
> Does not seem to work?
> 
> Screenshot 2019-09-10 07 52 42

What should be done:
- deploy functions
- clear database and browser cache

eddiejaoude:
> clear database

ok I will try

functions/src/project/delete-project.ts
kykyk: It is not required. This function return Promise<void>

functions/src/repository/create-repository.ts
kykyk: `async function` automatically wraps return value by `promise`

eddiejaoude:
Deleted the database but get the following error when adding a repository to the new created project

```
Unhandled error TypeError: Cannot read property 'fullName' of undefined`
```

eddiejaoude:
Interestingly it worked after the 5th attempt. I will keep investigating.

PR #1486, opened 2019-09-09
feat(repositories)#1123 parse webhooks from GitHub

### LINKED ISSUE #1123: Realtime events - parse webhooks from GitHub

kykyk:
closes #1123

Done:
- models for response of webhook
- logic of detect the event type
- parse events

Problem:
- update contributors (github cache https://developer.github.com/v3/repos/statistics/)


functions/src/mappers/github/pullRequest.mapper.ts
```diff
@@ -28,8 +25,8 @@ export interface GitHubPullRequestModel {
   owner: GitHubUserModel;
   assignees: GitHubUserModel[];
   reviewers: GitHubUserModel[];
-  createdOn: firestore.Timestamp;
-  updatedOn: firestore.Timestamp;
+  createdOn: string;
+  updatedOn: string;
```
eddiejaoude: Why did these change to `string`? I think Firestore has deprecated this in favour of `firestore.Timestamp`

kykyk: For now ‘createdOn’ and ‘updatedOn’ have a string type in our database. If we want to change it to Timestamp we need to create conversion function

PR #1491, opened 2019-09-10
feat(projects): #1478 added url , description and milestones

### LINKED ISSUE #1478: Repo rating part 2
Check repo for (if we have the data already, if not then ignore that rating)

- website url (100 % if exists, 0% if not)
- description (100% if exists, 0% if not)
- Milestone rating in the last 30 days (same as part 1)

webkhushboo:
closes #1478 

### Notes

- [ ] Added url , description and milestones in repo rating calculation

### DIFF
--- a/web/src/app/core/services/repository.service.ts
+++ b/web/src/app/core/services/repository.service.ts
@@ -59,7 +59,10 @@ export class RepositoryService {
     let rating: number = 0;
     const issuePoints: number = repo.issues.length > 0 ? this.getPoints(repo.issues[0].createdOn) : 0;
     const releasesPoints: number = repo.releases.length > 0 ? this.getPoints(repo.releases[0].createdOn) : 0;
-    rating = (issuePoints + releasesPoints) / 2;
+    const milestonesPoints: number = repo.milestones.length > 0 ? this.getPoints(new Date(repo.milestones[0].updatedAt)) : 0;
+    const urlPoints: number = repo.url ? 100 : 0;
+    const descriptionPoints: number = repo.description ? 100 : 0;
+    rating = (issuePoints + releasesPoints + milestonesPoints + urlPoints + descriptionPoints) / 5;
 
     return rating;
   }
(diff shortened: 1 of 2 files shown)


web/src/app/core/services/repository.service.ts
```diff
@@ -59,7 +59,10 @@ export class RepositoryService {
     let rating: number = 0;
     const issuePoints: number = repo.issues.length > 0 ? this.getPoints(repo.issues[0].createdOn) : 0;
     const releasesPoints: number = repo.releases.length > 0 ? this.getPoints(repo.releases[0].createdOn) : 0;
-    rating = (issuePoints + releasesPoints) / 2;
+    const milestonesPoints: number = repo.milestones.length > 0 ? this.getPoints(new Date(repo.milestones[0].updatedAt)) : 0;
+    const urlPoints: number = repo.url ? 100 : 0;
+    const descriptionPoints: number = repo.description ? 100 : 0;
+    rating = (issuePoints + releasesPoints + milestonesPoints + urlPoints + descriptionPoints) / 5;
```
eddiejaoude: Looks good 👍 . 

May be as we have many it might be cleaner to write a loop? Something like...

```typescript
const checks: number[] = [];
checks.push(repo.issues.length > 0 ? this.getPoints(repo.issues[0].createdOn) : 0);
checks.push(repo.releases.length > 0 ? this.getPoints(repo.releases[0].createdOn) : 0);		    
checks.push(repo.releases.length > 0 ? this.getPoints(repo.releases[0].createdOn) : 0);
checks.push(repo.milestones.length > 0 ? this.getPoints(new Date(repo.milestones[0].updatedAt)) : 0);
checks.push(repo.url ? 100 : 0);
checks.push(repo.description ? 100 : 0);

rating = checks.reduce((total, current) => total + current, 0) / checks.length
```

webkhushboo: do we need to add that in case if any of these fields doesn't exist

eddiejaoude: I think the flow is the same, it just seems easier for the last line, so we don't need to keep adding a `+ newitem` and then increment the divide. 

All that would need to be done for a new calculation is push to the `checks` array and the accumulation and divide would just figure itself out

webkhushboo: sure !! I have updated the method with above logic.

PR #1537, opened 2019-09-25
feat(repositories): #1533 repositories will show in descending order 

### LINKED ISSUE #1533: Order project repos by last updated date
On the project dashboard page, order repos by last updated date

webkhushboo:
closes #1533 

### Notes

- [ ] Show the repository in descending order

### DIFF
--- a/web/src/app/core/services/repository.service.ts
+++ b/web/src/app/core/services/repository.service.ts
@@ -34,7 +34,11 @@ export class RepositoryService {
     return this.activityService
       .start()
       .pipe(
-        switchMap(() => this.afs.collection<IRepository>('repositories').doc<IRepository>(uid).valueChanges()),
+        switchMap(() => this.afs.collection<IRepository>(
+          'repositories',
+          (ref: firebase.firestore.Query) => ref
+            .orderBy('updatedOn', 'desc')
+        ).doc<IRepository>(uid).valueChanges()),
         map((repository: IRepository) => new RepositoryModel(repository))
       );
   }


web/src/app/core/services/repository.service.ts
```diff
@@ -34,7 +34,11 @@ export class RepositoryService {
     return this.activityService
       .start()
       .pipe(
-        switchMap(() => this.afs.collection<IRepository>('repositories').doc<IRepository>(uid).valueChanges()),
+        switchMap(() => this.afs.collection<IRepository>(
+          'repositories',
+          (ref: firebase.firestore.Query) => ref
+            .orderBy('updatedOn', 'desc')
+        ).doc<IRepository>(uid).valueChanges()),
```
eddiejaoude: This is finding a single repository, ordering it doesn't make sense when there is only always 1 result?

I think we need to sort it here 

```html
    <div class="repository" *ngFor="let uid of project.repositories">
      <dashboard-repository [uid]="uid" [isAdmin]="isAdmin()"></dashboard-repository>
    </div>
```

https://github.com/DashboardHub/PipelineDashboard/blob/v0.11.9/web/src/app/projects/view/view.component.html#L75

webkhushboo: Here only uid of repositories are coming so we can't sort it here . Here we are just passing the uid of repository and showing it using another component. 
Either We need all repositories with the updatedOn date too or we have to create function to fetch all repositories then sort the uid and pass them here.

eddiejaoude: I think we can sort with a pipe on the array? https://angular.io/guide/pipes#custom-pipes

webkhushboo: But we are not getting the updatedOn field in the array . How we will sort ?

eddiejaoude: Oh that is a problem 😂🤦‍♂️

eddiejaoude: Ok this one needs some thought. May be we can update the array to be an array of objects with uid and updatedAt dates? I am not sure the knock on effect though, so would need looking into further

eddiejaoude: Lets park this ticket for now, I moved to blocked column as its not urgent

eddiejaoude:
Closing PR as the work is not simple and requires more data available

PR #1545, opened 2019-10-02
feat(project): #1541 added follow button in project dashboard

### LINKED ISSUE #1541: Follow project(s)
The ability to star/favourite/follow a public project

- On the project dashboard, have a follow button after the project name and display how many people are following the project (counter see below)
- Store project `uid` under the user doc collection store
   - `user.following: string[]`
- Page to list the projects the user follows -- This will be implemented in task #1547 
- Increment project counter `project.followers: number` (display on project page, see above)

Issue
- If project gets deleted? Show project as missing?

webkhushboo:
closes #1541 

### Notes

- [ ] Added follow button in project dashboard
- [ ] If user click on follow button then it changes to unfollow 
- [ ] It shows the count of no of followers as well


web/src/app/core/services/user.service.ts
```diff
@@ -75,4 +76,26 @@ export class UserService {
...
       );
   }
+
+  /**
+   * Update the following array inside users collection based upon the update flag
+   * @param userUid string
+   * @param projectUid string
+   * @param isUpdate flag if user uid has to remove or array in following array
+   */
+  public updateFollowings(userUid: string, projectUid: string, isUpdate: boolean): Observable<void> {
```
eddiejaoude: Is the `userUid` really needed here, will it not just be the logged in user?

web/src/app/projects/view/view.component.html
```diff
@@ -49,6 +49,12 @@
         <div class="project-body__header__content__item__pings" fxLayout="row" fxLayoutGap="5px">
           <mat-icon svgIcon="ping_icon"></mat-icon><span>{{ project.getTotalPings() }} | Total Pings</span>
         </div>
+        <div *ngIf= "!isFollower" (click)="incrementFollowers()" class="project-body__header__content__item" fxLayout="row" fxLayoutGap="5px">
```
eddiejaoude: I think the function name is confusing. The user is following the project, it does increment but that is not their action. I think `followProject` and `unfollowProject` would make more sense here

web/src/app/projects/view/view.component.ts
```diff
@@ -91,6 +94,31 @@ export class ViewProjectComponent implements OnInit, OnDestroy {
       .subscribe();
   }
 
+  /**
+   * Increase the followers count by 1 and update the followings in the users collection
+   */
+  incrementFollowers(): void {
+    this.projectService.updateFollowers(this.project.uid, 1).subscribe();
+    this.userService.updateFollowings(this.authService.profile.uid, this.project.uid, true).
```
eddiejaoude: The `profile.uid` could be used directly in the `userService` would not need to be included in and then passed through

web/src/app/projects/view/view.component.ts
```diff
@@ -91,6 +94,31 @@ export class ViewProjectComponent implements OnInit, OnDestroy {
...
+      pipe(
+        tap(() => this.setFollower())
+    )
+    .subscribe();
```
eddiejaoude: Empty `subscribe` does not seem quote right, can the `setFollower` be in the `subscribe`?

web/src/app/projects/view/view.component.ts
```diff
@@ -91,6 +94,31 @@ export class ViewProjectComponent implements OnInit, OnDestroy {
...
+    this.setFollower();
+  }
+
+  /**
+   * Decrease the followers count by 1 and update the followings in the users collection
+   */
+  decrementFollowers(): void {
```
eddiejaoude: Is this `public` or `private`?

web/src/app/shared/models/user.model.ts
```diff
@@ -7,6 +7,7 @@ export class UserModel {
   username: string;
   avatarUrl: string = '';
   url: string = '';
+  following?: string[] = [];
```
eddiejaoude: I don't think this is needed here? We use `ProfileModel` in our application, this is just for the authentication with GitHub OAuth - if I remember correctly

web/src/app/projects/view/view.component.html
```diff
@@ -49,6 +49,12 @@
...
+          <mat-icon svgIcon="follow_icon"></mat-icon><span>{{ project.followers }} | Follow </span>
+        </div>
+        <div *ngIf= "isFollower" (click)="decrementFollowers()" class="project-body__header__content__item" fxLayout="row" fxLayoutGap="5px">
+          <mat-icon svgIcon="unfollow_icon"></mat-icon><span>{{ project.followers }} | Unfollow </span>
+        </div>
```
eddiejaoude, line 61: This section only gets loaded if isAdmin(), other users can not follow the project

eddiejaoude:
When I `follow` and `unfollow` it seems to get an error and always decrease, so now I am on `-2`

Screenshot 2019-10-18 20 06 58

eddiejaoude:
The error was because I forgot to deploy the function 🤦‍♂

web/src/app/core/services/project.service.ts
```diff
@@ -243,4 +244,25 @@ export class ProjectService {
...
+  public updateFollowers(projectUid: string, counter: number): Observable<void> {
+    return this.activityService
+      .start()
+      .pipe(
+        switchMap(() => this.afs
+          .collection<IProject>('projects')
+          .doc<IProject>(projectUid)
+          .set(
+            {
+              followers: firestore.FieldValue.increment(counter),
```
eddiejaoude: Only the admin/owner of the project can write/update the project document. This will have to be done another way 🤔

eddiejaoude: When not the owner of a project, after allowing the follow/unfollow button appear for non-owners, when clicking...

Screenshot 2019-10-19 08 52 26

eddiejaoude: This might need to move to a callable function. I will have a look into it 🤓

web/src/app/projects/view/view.component.ts
```diff
@@ -123,6 +151,17 @@ export class ViewProjectComponent implements OnInit, OnDestroy {
     return this.project.isAdmin(this.authService.profile.uid);
   }
 
+  /**
+   * Set the followers flag true if user uid is present in the database else set false
+   */
+  public setFollower(): void {
+    if (this.authService.profile.following.includes(this.project.uid)) {
```
eddiejaoude: User might be a guest and not logged in, `this.authService.profile.following` can be `undefined`

eddiejaoude: Fixed by putting a default on the `ProfileModel` 

```typescript
following?: string[] = [];
```

web/src/app/projects/view/view.component.ts
```diff
@@ -106,6 +109,31 @@ export class ViewProjectComponent implements OnInit, OnDestroy {
       .subscribe();
   }
 
+  /**
+   * Increase the followers count by 1 and update the followings in the users collection
+   */
+  incrementFollowers(): void {
+    this.projectService.updateFollowers(this.project.uid, 1).subscribe();
```
eddiejaoude: These 2 service calls with the empty subscribes could be combined

web/src/app/projects/view/view.component.ts
```diff
@@ -123,6 +151,17 @@ export class ViewProjectComponent implements OnInit, OnDestroy {
     return this.project.isAdmin(this.authService.profile.uid);
   }
 
+  /**
+   * Set the followers flag true if user uid is present in the database else set false
+   */
+  public setFollower(): void {
```
eddiejaoude: I think `isFollower()` would make more sense, then this can be called where needed

eddiejaoude (approved):

web/src/app/projects/view/view.component.html
webkhushboo: Cool

web/src/app/core/services/user.service.ts
webkhushboo: Yes it should be logged in user

web/src/app/projects/view/view.component.ts
webkhushboo: ok

web/src/app/shared/models/user.model.ts
webkhushboo: ok got it

web/src/app/projects/view/view.component.ts
webkhushboo: public

PR #1553, opened 2019-10-05
feat(analytics): #1550 added google analytics tracking for router events

### LINKED ISSUE #1550: Google analytics for App and Website
I added google analytics to `index.html`, but apparently there is a better way to do it for Angular, to capture which pages etc

Please also do the same for this repo https://github.com/DashboardHub/Website

webkhushboo:
closes #1550 

### Notes

- Added google analytics tracking code for router events

### DIFF
--- a/web/src/environments/environment.ts
+++ b/web/src/environments/environment.ts
@@ -13,5 +13,5 @@ export const environment: Config = {
     messagingSenderId: '',
   },
 
-  tracking: 'UA-60804146-5',
+  tracking: '',
 };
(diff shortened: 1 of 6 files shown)


web/src/environments/environment.ts
```diff
@@ -12,4 +12,6 @@ export const environment: Config = {
     storageBucket: '',
     messagingSenderId: '',
   },
+
+  tracking: 'UA-60804146-5',
```
eddiejaoude: I think this file can have an empty string?

eddiejaoude (approved):

web/src/environments/environment.ts
webkhushboo: done

PR #1598, opened 2019-11-04
feat(projects): #1594 filtered the public projects having > 1 repo

### LINKED ISSUE #1594: On homepage, only show projects in the public project list if they have at least 1 repository

webkhushboo:
closes #1594 

### Notes

- [ ] UI: Display public projects if having more than 1 repository

image

### DIFF
--- a/web/src/app/shared/components/private-public-project/private-public-project.component.ts
+++ b/web/src/app/shared/components/private-public-project/private-public-project.component.ts
@@ -41,7 +41,7 @@ export class PrivatePublicProjectComponent implements OnInit, OnDestroy {
       this.projectSubscription = this.projectService
         .findPublicProjects()
         .subscribe((projects: IProject[]) => this.projects = projects.map((project: IProject) => new ProjectModel(project))
-          .filter((project: ProjectModel) => project.repositories.length > 1));
+          .filter((project: ProjectModel) => project.repositories.length > 0));
     } else {
       this.showTitle = true;
       this.projectSubscription = this.projectService
--- a/web/src/app/shared/components/private-public-project/private-public-project.component.ts
+++ b/web/src/app/shared/components/private-public-project/private-public-project.component.ts
@@ -40,7 +40,8 @@ export class PrivatePublicProjectComponent implements OnInit, OnDestroy {
     if (this.router.url === '/') {
       this.projectSubscription = this.projectService
         .findPublicProjects()
-        .subscribe((projects: IProject[]) => this.projects = projects.map((project: IProject) => new ProjectModel(project)));
+        .subscribe((projects: IProject[]) => this.projects = projects.map((project: IProject) => new ProjectModel(project))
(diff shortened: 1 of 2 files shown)


webkhushboo:
@eddiejaoude 

I have tagged the wrong issue in the commit https://github.com/DashboardHub/PipelineDashboard/pull/1598/commits/7ce11110ac73a6be4709ed70843fb20c950e6136

Please squash commit while merge. if possible . It would be really helpful

web/src/app/shared/components/private-public-project/private-public-project.component.ts
```diff
@@ -40,7 +40,8 @@ export class PrivatePublicProjectComponent implements OnInit, OnDestroy {
     if (this.router.url === '/') {
       this.projectSubscription = this.projectService
         .findPublicProjects()
-        .subscribe((projects: IProject[]) => this.projects = projects.map((project: IProject) => new ProjectModel(project)));
+        .subscribe((projects: IProject[]) => this.projects = projects.map((project: IProject) => new ProjectModel(project))
+          .filter((project: ProjectModel) => project.repositories.length > 1));
```
eddiejaoude: `> 1`, that will be `2` or more? We need it to be `1` or more, so `> 0` would be better?

web/src/app/shared/components/private-public-project/private-public-project.component.ts
```diff
@@ -40,7 +40,8 @@ export class PrivatePublicProjectComponent implements OnInit, OnDestroy {
     if (this.router.url === '/') {
       this.projectSubscription = this.projectService
         .findPublicProjects()
-        .subscribe((projects: IProject[]) => this.projects = projects.map((project: IProject) => new ProjectModel(project)));
+        .subscribe((projects: IProject[]) => this.projects = projects.map((project: IProject) => new ProjectModel(project))
+          .filter((project: ProjectModel) => project.repositories.length > 0));
```
eddiejaoude, line 44: When the list get longer in the future it would be more efficient to get the database to do the filtering if possible.

Currently the query limits to 10 results, if we remove items after the query we could end up with less than 10

```typescript
ref.where('type', '==', 'public')
              .orderBy('updatedOn', 'desc').limit(10)
```

eddiejaoude: It does not look like it is possible to do a `repositories.length > 0` query in Firestore. We would need to store the `repositories` count to do this.

Best to store the repository length, so we can do a query to Firestore.

webkhushboo: Okay Do I have to make this change in this issue ? or we can do it in another ?

eddiejaoude: I think do it in this PR, as this PR is tiny

webkhushboo: okay I will do it here

PR #1602, opened 2019-11-08
feat(automation): #1596 added automation scripts for homepage titles

### LINKED ISSUE #1596: Automated testing part 1 - setup + homepage
I am keen to refactor the application, however we need to have automated testing with `Cypress + Cucumber`, I know in the past we tried to mock out firebase and had issues, now lets try with no mocking, use the current environment (`eddie`, `khush`, and CI to use `dev`)

- UI
- http functions (webhooks)

This can be split into multiple PRs, my initial thinking for this issue is..

1. setup cypress + cucumber
2. test the homepage on `localhost:4200`
   - check guest user at the top right
   - check for `Application statistics` title
   - check for `Active Users`  title
   - check for `Popular projects` title
   - check for `Public Projects` title

webkhushboo:
closes #1596 

### Notes

- [ ] Automation testing scripts for homepage titles validation


web/test/step_definitions/browser.js
```diff
@@ -1,11 +1,16 @@
...
 Given(/^the "([^"]*)" page is open$/, (path) => {
     cy.visit(`${url}/${path}`);
 });
 
 Then(/^the title on the page says "([^"]*)"$/, (check) => {
-    cy.title().should('include', check);
+  cy.title().should('include', check);
 });
+
+And(/^the text "([^"]*)" is in the element "([^"]*)"$/, (text, element) => {
```
eddiejaoude: There is no `And` import, this should be `Then`, but in the `.feature` file, `And` can be used

web/test/step_definitions/browser.js
```diff
@@ -1,11 +1,16 @@
 import { Given, Then } from 'cypress-cucumber-preprocessor/steps';
 
-const url = 'http://localhost:8080';
+const url = 'http://localhost:4200';
 
 Given(/^the "([^"]*)" page is open$/, (path) => {
     cy.visit(`${url}/${path}`);
 });
 
 Then(/^the title on the page says "([^"]*)"$/, (check) => {
```
eddiejaoude: Can `check` be `text` to be consistent with the other step def

webkhushboo: Actually its verifying the title on  the page not the element . It has different step definition checking `cy.title()` which Get the `document.title` property of the page that is currently active.

web/test/step_definitions/browser.js
webkhushboo: okay . I will update this.

web/test/step_definitions/browser.js
eddiejaoude: I mean for the variable name, it is checking for text, so I think it would be more consistent to have use `text` instead of `check` for the variable name

webkhushboo: okay gotcha !! will fix it too :) Thank you for clarifying 👍

eddiejaoude (approved):

Issue #1603, opened 2019-11-12
Delete repository subcollection


kykyk:
- add function with transaction for delete the collection
- delete repo subcollection `statuses`

PR #1604, opened 2019-11-12
fix(repository): #1603 delete repo subcollection

kykyk:
Done:

- add function with transaction for delete the collection
- delete repo subcollection statuses

### DIFF
--- a/functions/package.json
+++ b/functions/package.json
@@ -16,6 +16,7 @@
   "dependencies": {
     "@google-cloud/logging-winston": "^1.0.0",
     "cors": "^2.8.5",
+    "firebase": "^7.3.0",
     "firebase-admin": "^8.2.0",
     "firebase-functions": "^3.1.0",
     "request": "^2.88.0",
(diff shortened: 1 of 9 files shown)


eddiejaoude:
Sorry for my slow reply.

After testing it appears that the documents in the sub collection `statuses` do get deleted but the sub collection itself remains (although empty). I think the empty sub collection needs to also needs to get deleted

Screenshot 2019-11-18 07 58 48

kykyk:
> Sorry for my slow reply.
> 
> After testing it appears that the documents in the sub collection `statuses` do get deleted but the sub collection itself remains (although empty). I think the empty sub collection needs to also needs to get deleted
> 
> Screenshot 2019-11-18 07 58 48

Could you provide me a chain of action you did for reproducing this bug?

eddiejaoude:
> Could you provide me a chain of action you did for reproducing this bug?

Sure...
1. create project
2. add repo to project
3. create PR in repo on GitHub (`statuses` sub collection now has an entry)
4. delete project

All documents in the `statuses` sub collection were deleted correctly but the sub collection itself remains

kykyk:
I checked a problem
It is not bug, the empty subcollection in firebase automatically delete through sometime

PR #1607, opened 2019-11-14
Issue 1606: feat(testing): saving project json in db and validate 

### LINKED ISSUE #1606: Automated test part 3 - public projects
- Insert a public project into the DB (currently we only do fields, we need a way to insert a whole json file) - also put the date in the future so it appears at the top of the list
- Insert a private project and check it does not appear in the list

webkhushboo:
closes #1606

### Notes

- [ ] Saving public project json in db 
- [ ] Validate public project title on UI
- [ ] Saving private project json in db
- [ ] Validate private project title should not be visible on UI


web/test/data/public-project.json
```diff
@@ -0,0 +1,6 @@
+{
+  "type": "public",
+  "title": "Demo public project",
+  "description": "Demo public project description added",
+  "url": "https://www.pipelinedashboard.io"
+}
```
eddiejaoude, line 6: I was initially thinking of a complete project, but I think this is better, it tests that the application can handle minimum data. I will create new tickets for more details data like repositories etc

web/test/integration/project.feature
```diff
@@ -0,0 +1,11 @@
+Feature:  Save the project json in db
```
eddiejaoude: This does not describe the epic/result we are expecting, I would I suggest something like `Display projects on the homepage`

web/test/integration/project.feature
```diff
@@ -0,0 +1,11 @@
+Feature:  Save the project json in db 
+
+Scenario: Ensure the create public project
```
eddiejaoude: The test here is not the create, that is the step def, the test is the reading. I would suggest something like `List public project on the homepage`

web/test/plugins/index.js
```diff
@@ -20,4 +20,12 @@ module.exports = (on, config) => {
     }
   });
 
+  on('task', {
+    'db:update:project': (params) => {
```
eddiejaoude: I suggest `db:project:save`

web/test/data/public-project.json
webkhushboo: okay

web/test/integration/project.feature
webkhushboo: okay will change the text

web/test/integration/project.feature
webkhushboo: okay will update

web/test/plugins/index.js
webkhushboo: okay will do

eddiejaoude (approved):

PR #1619, opened 2019-12-01
feat(testing): #1618 reset the db after each test run

### LINKED ISSUE #1618: Reset the DB before each test run
Cypress has hooks to run actions before each test file.

Only delete the test documents in the collections, check the `uid` for `test-` then delete

webkhushboo:
closes #1618 

### Notes

- [ ] Reset the DB after each test run in cypress hooks


web/test/support/index.js
```diff
@@ -1,6 +1,8 @@
 import './commands';
 
 before(() => {
+  cy.task('db:delete', {collection: 'projects', subCollection: 'pings'})
+  cy.task('db:delete', {collection: 'repositories', subCollection: 'statuses'})
```
eddiejaoude: It might be cleaner if the `collection` and `subcollections` (array of multiples) where all in a single object, and all got deleted, what do you think?

```typescript
{
    collection: 'repositories', 
    subCollections: [ 'statuses' ]
}
```

web/test/plugins/index.js
```diff
@@ -61,6 +61,14 @@ module.exports = (on, config) => {
         querySnapshot.forEach(function (doc) {
           if ((doc.id).startsWith("test-")) {
             db.collection(params.collection).doc(doc.id).delete();
+            if (params.subCollection) {
+              db.collection(`${params.collection} / ${doc.id} / ${params.subCollection}`).get().then(
+                function (subCollectionSnapshot) {
```
eddiejaoude: Can we use es6 functions

web/test/plugins/index.js
```diff
@@ -55,4 +55,16 @@ module.exports = (on, config) => {
...
+  on('task', {
+    'db:delete': (params) => {
+      db.collection(params.collection).get().then(function (querySnapshot) {
+        querySnapshot.forEach(function (doc) {
+          if ((doc.id).startsWith("test-")) {
+            db.collection(params.collection).doc(doc.id).delete();
+          }
+        });
+      });
+      return null;
```
eddiejaoude: Is the `return` needed?

webkhushboo: Yes If I  do not return anything , it throws an error "Task should return promise or value"

eddiejaoude: I think it would be best then to not use `{}` to open a function and just return the chain, something like...

```typescript
'db:delete': (params) => db.collection(params.collection).get()
     .then((querySnapshot) => querySnapshot.forEach((doc) => {
           if ((doc.id).startsWith("test-")) {
             db.collection(params.collection).doc(doc.id).delete();
           }
         });
       });
```

webkhushboo: Yes , I will make these changes

web/test/plugins/index.js
webkhushboo: Will do

web/test/integration/homepage/projects.feature
```diff
@@ -28,8 +28,8 @@ Feature:  Display public projects on the homepage
     And the text "1 second ago" is in the element ".mat-column-lastDate"
 
   Scenario: Check private projects are not displayed on the homepage
-    Given there is a document "test-project-private-minimum" with the js "projects/project-private-minimum" in collection "projects"
+    When the "/" page is open
```
eddiejaoude: Data going in should be a `Given` and come first

web/src/app/shared/models/profile.model.ts
```diff
@@ -33,5 +33,5 @@ export class ProfileModel {
   repositories?: RepositoriesModel = new RepositoriesModel();
   activity?: ActivityModel[] = [];
   following?: string[] = [];
-  isAdmin?: Boolean = false;
+  isAdmin: boolean = false;
```
webkhushboo: Making it required creates an issue and it throws an error
`Type 'ProfileModel' is not assignable to type '{ uid: string; username: string; name: string; email: string; phone: string; avatarUrl: string; oauth: { githubToken: any; token: string; expirationTime: string; authTime: string; issuedAtTime: string; signInProvider: string; claims: { ...; }; }; emailVerified: boolean; creationTime: string; lastSignInTime: string; }'.
    Property 'oauth' is optional in type 'ProfileModel' but required in type '{ uid: string; username: string; name: string; email: string;
phone: string; avatarUrl: string; oauth: { githubToken: any; token: string; expirationTime: string; authTime: string; issuedAtTime: string;
signInProvider: string; claims: { ...; }; }; emailVerified: boolean; creationTime: string; lastSignInTime: string; }'.`

eddiejaoude:
If all data from `.js` files can be put in the feature files, then we can merge this PR

eddiejaoude (approved):

